PR #3, opened 2019-08-22
Change URL used to download product images for seed data.

mttdffy:


bencrouse:
Weird, I thought the header specifying to accept raw took care of this?

mttdffy:
> Weird, I thought the header specifying to accept raw took care of this?

Pretty sure we could get rid of the header code if we wanted. I think its more about the api being rate limited and the raw url isn't. Just a guess though based on observed behavior.

PR #4, opened 2019-08-22
Add Generic Activity Partials

tubbo:
Empty results were still being seen in the trash when a model that
doesn't explicitly have an activity partial defined is encountered. This
is due to the `render_activity_entry` helper rescuing an
`ActionView::TemplateError` to return a blank string. To resolve this
issue, models that are tracked by `Mongoid::AuditLog`, without an
explicit activity partial defined will be rendered using a generic
partial, showing the class name and ID of the audited model, as
something to render in the listing so that pages of blank results
aren't shown.


admin/app/helpers/workarea/admin/activities_helper.rb
```diff
@@ -4,10 +4,11 @@ module ActivitiesHelper
       def render_activity_entry(entry)
         model_key = ActiveModel::Naming.param_key(entry.audited_type.constantize)
         partial_name = "#{model_key}_#{entry.action}"
-        render "workarea/admin/activities/#{partial_name}", entry: entry
 
+        render "workarea/admin/activities/#{partial_name}", entry: entry
       rescue ActionView::MissingTemplate
-        '' # we don't want to render anything if the partial is missing
+        return '' unless entry.action == :destroy
```
bencrouse: This seems weird... why do we only want generic activity for a destroy?

PR #5, opened 2019-08-22
Prevent Deleting Category Referenced In Content

tubbo:
The Category Summary content block includes a reference to the category
by its ID, which can be dereferenced by deleting the category with no
warning to the content block. As a result, any page using the category
summary block with the deleted category ID will throw an error when
attempting to find the category by its ID. To fix this, prevent deletion
of the category through admin if a content block is referencing the
category ID.


bencrouse:
We don't really do this "foreign key" constraining anywhere else, I'd prefer not to do it in this one specific case. The most Workarea thing to do is have this fallback gracefully if the category can't be found.

PR #7, opened 2019-08-22
Fix Incorrect Currency in Mongoid Money Types

tubbo:
Workarea's default values for the Money fields in `Pricing::Override`
didn't previously change currency when `Money.default_currency` is
re-configured in process (like in the case of a multi-site application
with multiple currencies). Ensure that the correct currency is used by
using an Integer type as the default, which will get converted into a
Money type at runtime.

### DIFF
--- a/core/app/models/workarea/pricing/override.rb
+++ b/core/app/models/workarea/pricing/override.rb
@@ -6,8 +6,8 @@ module Workarea
       # This will be the order id
       field :_id, type: String, default: -> { BSON::ObjectId.new.to_s }
       field :created_by_id, type: String
-      field :subtotal_adjustment, type: Money, default: 0.to_m
-      field :shipping_adjustment, type: Money, default: 0.to_m
+      field :subtotal_adjustment, type: Money, default: 0
+      field :shipping_adjustment, type: Money, default: 0
       field :item_prices, type: Hash, default: {}
 
       def has_adjustments?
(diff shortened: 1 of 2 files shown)


core/app/models/workarea/pricing/override.rb
```diff
@@ -6,8 +6,8 @@ class Override
       # This will be the order id
       field :_id, type: String, default: -> { BSON::ObjectId.new.to_s }
       field :created_by_id, type: String
-      field :subtotal_adjustment, type: Money, default: 0.to_m
-      field :shipping_adjustment, type: Money, default: 0.to_m
+      field :subtotal_adjustment, type: Money, default: -> { 0.to_m }
```
bencrouse: A nicer, simpler way to solve this is to just leave it as `0` (omit the `.to_m`). We do this in `Order`: https://github.com/workarea-commerce/workarea/blob/master/core/app/models/workarea/order.rb#L24

Here's a demo:
```
irb(main):006:0> Money.default_currency = Money::Currency.new('AUD')
=> #<Money::Currency id: aud, priority: 4, symbol_first: true, thousands_separator: ,, html_entity: $, decimal_mark: ., name: Australian Dollar, symbol: $, subunit_to_unit: 100, exponent: 2, iso_code: AUD, iso_numeric: 036, subunit: Cent, smallest_denomination: 5>
irb(main):007:0> Workarea::Order.new.total_price
=> #<Money fractional:0 currency:AUD>
```

tubbo: i'll do this in gift-cards as well, where the problem was first reported.

PR #8, opened 2019-08-26
Update README.md

btarbell:
Added section showing usage of Workarea in physical retail format.


bencrouse:
Going to close since a single link isn't very compelling for a list and I don't think linking to news articles is a great look for READMEs in general.

PR #16, opened 2019-08-27
Remove Search Autocomplete

meowsus:
This functionality will one day live on in https://github.com/workarea-commerce/workarea-classic-search-autocomplete


meowsus:
> Few touch ups, and `core/app/assets/javascripts/workarea/core/templates/ui_menu_item.jst.ejs` looks like it's used by the admin, right?

@bencrouse Admin uses it's own version of that template.

PR #20, opened 2019-08-28
Use Current Host For Direct Upload CORS Headers

### LINKED ISSUE #19: Direct Uploads Fail Locally
**Describe the bug**
When directly uploading assets, either on `/admin/content_assets` or in a content block itself, and the S3 bucket is configured properly, assets can still fail to upload if the `Workarea.config.host` does not match 

**To Reproduce**
1. Configure your app with S3 secrets and a `config.host` like `example.com`. 
2. Go to `/admin/content_assets` and try to upload an asset
3. You'll get a "Something Went Wrong" message, and CORS warnings in the JS console

**Expected behavior**
The asset should upload successfully.

- Workarea v3.4.14

tubbo:
Direct uploads can fail locally if your `Workarea.config.host` is not
set to the domain you are currently using in the browser. To prevent
this, instead of reading from the configuration when ensuring CORS
headers on the S3 bucket, use the `ActionDispatch::Request` from the
controller to determine the correct URL for CORS in this instance.
Addresses a problem whereby changing the domain (either accidentally or
on-purpose) causes direct uploads to fail, since it can't create the
proper CORS headers needed to transmit files into the bucket directly.

Fixes #19

### DIFF
--- a/admin/app/controllers/workarea/admin/content_assets_controller.rb
+++ b/admin/app/controllers/workarea/admin/content_assets_controller.rb
@@ -7,7 +7,9 @@ module Workarea
     after_action :track_index_filters, only: :index
 
     def index
-      DirectUpload.ensure_cors! if Configuration::S3.configured? && !request.xhr?
+      if Configuration::S3.configured? && !request.xhr?
+        DirectUpload.ensure_cors!(request)
+      end
 
       search = Search::AdminAssets.new(params)
       @search = Admin::SearchViewModel.new(search, view_model_options)
--- a/admin/app/controllers/workarea/admin/content_assets_controller.rb
+++ b/admin/app/controllers/workarea/admin/content_assets_controller.rb
@@ -8,7 +8,7 @@ module Workarea
 
     def index
       if Configuration::S3.configured? && !request.xhr?
-        DirectUpload.ensure_cors!(request)
+        DirectUpload.ensure_cors!(request.url)
       end
 
       search = Search::AdminAssets.new(params)
--- a/admin/app/controllers/workarea/admin/direct_uploads_controller.rb
+++ b/admin/app/controllers/workarea/admin/direct_uploads_controller.rb
@@ -2,7 +2,7 @@ module Workarea
   module Admin
     class DirectUploadsController < Admin::ApplicationController
       def product_images
-        DirectUpload.ensure_cors! if Configuration::S3.configured?
+        DirectUpload.ensure_cors!(request.host) if Configuration::S3.configured?
       end
 
       def new
--- a/admin/app/controllers/workarea/admin/direct_uploads_controller.rb
+++ b/admin/app/controllers/workarea/admin/direct_uploads_controller.rb
@@ -2,7 +2,7 @@ module Workarea
   module Admin
     class DirectUploadsController < Admin::ApplicationController
       def product_images
-        DirectUpload.ensure_cors!(request) if Configuration::S3.configured?
+        DirectUpload.ensure_cors!(request.url) if Configuration::S3.configured?
       end
 
       def new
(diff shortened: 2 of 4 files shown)


admin/app/controllers/workarea/admin/content_assets_controller.rb
```diff
@@ -7,7 +7,9 @@ class Admin::ContentAssetsController < Admin::ApplicationController
     after_action :track_index_filters, only: :index
 
     def index
-      DirectUpload.ensure_cors! if Configuration::S3.configured? && !request.xhr?
+      if Configuration::S3.configured? && !request.xhr?
+        DirectUpload.ensure_cors!(request)
```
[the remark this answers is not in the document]
tubbo: I need the port sometimes.

admin/app/controllers/workarea/admin/direct_uploads_controller.rb
```diff
@@ -2,7 +2,7 @@ module Workarea
   module Admin
     class DirectUploadsController < Admin::ApplicationController
       def product_images
-        DirectUpload.ensure_cors! if Configuration::S3.configured?
+        DirectUpload.ensure_cors!(request.host) if Configuration::S3.configured?
```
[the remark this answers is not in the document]
tubbo: yup, that was a mistake

core/test/services/workarea/direct_upload_test.rb
```diff
@@ -46,6 +46,64 @@ def test_validation
       refute(direct_upload.valid?)
     end
 
+    def test_ensure_cors
+      original_host = Workarea.config.host
+      Workarea.config.host = 'test.host'
+      dev_request = mock(
```
bencrouse: Can we use `ActionDispatch::TestRequest` since Rails provides that for this kind of usecase? https://api.rubyonrails.org/v5.1/classes/ActionDispatch/TestRequest.html

core/test/services/workarea/direct_upload_test.rb
```diff
@@ -46,6 +46,64 @@ def test_validation
...
+            'AllowedHeader' => '*'
+          }
+        ]
+      ).returns(true)
+
+
+      assert(DirectUpload.ensure_cors!)
+      assert(DirectUpload.ensure_cors!(dev_request))
+      assert(DirectUpload.ensure_cors!(prod_request))
+    ensure
```
bencrouse: Don't forget about `Workarea.with_config do |config|`!

core/app/services/workarea/direct_upload.rb
```diff
@@ -2,17 +2,25 @@ module Workarea
   class DirectUpload
     class InvalidTypeError < RuntimeError; end
 
-    def self.ensure_cors!
-      url = "http#{Rails.application.config.force_ssl ? 's' : ''}://"
-      url += Workarea.config.host
+    def self.ensure_cors!(request = nil)
```
bencrouse: Should we just make the url a required arg if it's going to be used everywhere? Then we could clean up that `if` clutter below! Something like: `DirectUpload.ensure_cors!(request.url)`. `request.url` gives me `http://localhost:3000`, which is everything we'd need here. That would also alleviate the need to mock in the test.

tubbo: I'm down with that.

core/app/services/workarea/direct_upload.rb
```diff
@@ -2,15 +2,14 @@ module Workarea
   class DirectUpload
     class InvalidTypeError < RuntimeError; end
 
-    def self.ensure_cors!
-      url = "http#{Rails.application.config.force_ssl ? 's' : ''}://"
-      url += Workarea.config.host
+    def self.ensure_cors!(url)
+      uri = URI.parse(url)
```
bencrouse: We can just get rid of this right? No need for just the host below, might as well use the full URL as the ID in the config.

meowsus:
[Direct Uploads Fail Locally](https://trello.com/c/jjQp5hT3/3-direct-uploads-fail-locally)

Issue #31, opened 2019-09-05
check_mongo_notable_scan fails when adding authentication to mongodb


mcfearsome:
**Describe the bug**
`Workare::Configuration::Mongoid.indexes_enforced?` does not support mongodb authentication as is.

**To Reproduce**
Add authentication to mongodb, configure mongoid.yml for authentication, and start the application

**Expected behavior**
Application to start up without error

**Workarea Setup (please complete the following information):**
 - Workarea Version: 3.3.32

**Additional context**
`check_mongo_notable_scan` checks for `!Rails.env.test?`, the described issue could be avoided by checking for `Rails.env.development?` instead.  Since tests are never run against a qa/staging/production mongo cluster, it is not possible for them to enter the state we are checking (without manual intervention) and so the need to check for it in these environments is diminished.

PR #32, opened 2019-09-06
Only check notablescan in development

mcfearsome:
#31

### DIFF
--- a/core/lib/workarea/configuration/mongoid.rb
+++ b/core/lib/workarea/configuration/mongoid.rb
@@ -13,10 +13,7 @@ module Workarea
       end
 
       def indexes_enforced?
-        servers = ::Mongoid::Clients.default.cluster.servers
-        addresses = servers.map(&:address).map(&:to_s)
-
-        client = Mongo::Client.new(addresses, database: 'admin')
+        client = ::Mongoid::Clients.default.use('admin')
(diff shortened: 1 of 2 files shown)


bencrouse:
So is the goal with this one to prevent admin db access or to fix the fact that the client we use for the notablescan check doesn't use the same config as the Mongoid default? I ask because this might be a better fix: https://github.com/workarea-commerce/workarea/commit/9d184e7193b844c58a2cf436d9e218eadb8655e2

But only if the goal isn't to prevent admin db access. Maybe I just merge that into this branch and we get both fixes?

mcfearsome:
When I looked at what was being checked, it seemed to me it wasn't required unless you were running tests against a specific environments mongo instance.  Perhaps this is only relevant for our way of doing things, but the staging/production mongo instances will never have tests ran against them and so this fix seemed the simplest course of action.

bencrouse:
Yeah, I'm trying to understand the goal a bit deeper to see if that other commit is actually a better (deeper) fix. But that's dependent on whether we think the app should have access to the admin database at all. If we do think the app should have admin db access, then the other fix would be preferable since then if notablescan got turned on in any non-test env, you'd still get a warning.

bencrouse:
Thanks @mcfearsome !

PR #46, opened 2019-09-10
Parse URL When Ensuring CORS for Direct Uploads

### LINKED ISSUE #19: Direct Uploads Fail Locally

tubbo:
The `request.url` returns the full URL, with path included. This isn't
valid for a CORS header, which needs just the scheme, host, and port if
it's non-standard. Update the `DirectUpload.ensure_cors!` method to
parse out those pieces of the URL and re-assemble it for the CORS
header and ID.

No changelog

Fixes #19

### DIFF
--- a/core/app/services/workarea/direct_upload.rb
+++ b/core/app/services/workarea/direct_upload.rb
@@ -2,7 +2,11 @@ module Workarea
   class DirectUpload
     class InvalidTypeError < RuntimeError; end
 
-    def self.ensure_cors!(url)
+    def self.ensure_cors!(request_url)
+      uri = URI.parse(request_url)
+      url = "#{uri.scheme}://#{uri.host}"
+      url += ":#{uri.port}" unless uri.port.in? [80, 443]
+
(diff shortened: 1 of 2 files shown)


bencrouse:
Why does the port being "non-standard" matter?

tubbo:
> Why does the port being "non-standard" matter?

Because in a browser, `:80` and `:443` are assumed when you use `http://` and `https://`, respectively. If I didn't filter out the port, the URL would be something like `http://store.test:80` and that would not create the right CORS configuration for the bucket.

PR #48, opened 2019-09-10
Indexing performance

### LINKED ISSUE #37: Improve product indexing performance

mttdffy:
resolves #37

### DIFF
--- a/core/app/middleware/workarea/query_cache_middleware.rb
+++ b/core/app/middleware/workarea/query_cache_middleware.rb
@@ -1,0 +1,55 @@
+module Workarea
+  # Execute sidekiq jobs with query caches enabled. Each job can specify one or
+  # both query_caches to use via `query_cache`, `mongoid_query_cache`, or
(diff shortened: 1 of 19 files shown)


core/lib/workarea/elasticsearch/query_cache.rb
```diff
@@ -102,6 +102,33 @@ def call(env)
         end
       end
 
+      # Execute sidekiq jobs with query cache enabled.
+      # Only enabled for workers that set use_elasticsearch_query_cache to true.
+      #
+      # @return [ Object ] The result of the call.
+      #
+      class SidekiqMiddleware
```
tubbo: I kinda feel like this should be in **app/middleware** but i might be in the minority :P

mttdffy: The other middleware for query cache is in this file to, thought I'd be consistent.

tubbo: now that’s what I call consistent.

PR #49, opened 2019-09-10
Update admin configuration options

mttdffy:
* Add `allow_blank` option to field
* Require a value on fields that do not `allow_blank`
* Set field to default value if submitted blank when `allow_blank` is 
`false`.
* Remove `search_facet_default_sort` as admin configuration field to 
prevent confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ This will run a script that does the following:
 * seed the database
 * start the application server
 
-It requires you have Docker installed and running. Once complete, you can visit `http://localhost:3000` to view your storefront app. For the Workarea Admin, you can visit 'http://localhost:3000/admin'. The seed data provides an admin user with an email/password of `user@workarea.com/W3bl1nc!`.
+It requires you have Docker installed and running. Once complete, you can visit `http://localhost:3000` to view your storefront app. For the Workarea Admin, you can visit 'http://localhost:3000/admin'. The seed data provides an admin user with an email/password of `user@workarea.com/w0rkArea!`.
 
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -80,7 +80,7 @@ This will run a script that does the following:
 * seed the database
 * start the application server
 
-It requires you have Docker installed and running. Once complete, you can visit `http://localhost:3000` to view your storefront app. For the Workarea Admin, you can visit 'http://localhost:3000/admin'. The seed data provides an admin user with an email/password of `user@workarea.com/W3bl1nc!`.
+It requires you have Docker installed and running. Once complete, you can visit `http://localhost:3000` to view your storefront app. For the Workarea Admin, you can visit 'http://localhost:3000/admin'. The seed data provides an admin user with an email/password of `user@workarea.com/w0rkArea!`.
```
bencrouse: Nice

core/config/initializers/00_configuration.rb
```diff
@@ -131,16 +131,6 @@
         in the default config above.
       ).squish
 
-    field 'Search Facet Default Sort',
```
bencrouse: Why are we dropping this? Seemed useful.

mttdffy: The default sort only applies to facets that do that have a sort specified. This caused confusion for Jeff, and would likely for any admin as the facet-specific sorts are not admin configurable due to procs and classes being allowed as options. Given you can't see or change the specific ones, the generic default seems to do nothing out of the box as it does not affect the default facet types, which all have a sort specified in the static config.

PR #54, opened 2019-09-11
Ensure test only asserts product details for product system test

### LINKED ISSUE #42: Fix test_showing_a_specific_variant_by_sku

mttdffy:
fixes #42

I cannot confirm, but my suspicion is that recently viewed may be the culprit here. I've adjusted the test to only look at the product details specifically.


bencrouse:
Ok, I'm going to re-run the build a few times to try to verify before merging.

bencrouse:
Wow, apparently you can't re-run a successful build? What a world. Running one of the failed ones...

PR #55, opened 2019-09-11
Fix Wrapping `nil` Values in Product View Model

### LINKED ISSUE #33: Unhandled exception in admin when viewing a product
Workarea 3.4.11

* Product matches a category product_rule
* Category is not active.
* Admin views product and the "Categories" card attempts to reference an inactive category. 

Error: ```ActionView::Template::Error: undefined method `id' for nil:NilClass```

This test demonstrates the error:

```
require 'test_helper'

module Workarea
  module Admin
    class ProductCardsIntegrationTest < Workarea::IntegrationTest
      include Admin::IntegrationTest

      setup do
        rule = {
          name: 'price',
          operator: 'less_than',
          value: '10'
        }
        create_category(product_rules: [rule], active: false)
        @product = create_product
      end

      def test_view_product_cards
        get admin.catalog_product_path(@product)
      end
    end
  end
end
```

tubbo:
`ProductViewModel#default_category` now protects against a `nil` value
for the default category before passing its value into a view model.

My first approach was to try and change `ApplicationViewModel.wrap` to protect against `nil` values across the board, but this broke some stuff in `ShippingCarrierViewModel`, which depends on `model` being able to be `nil` and kinda acts as a null object...

Fixes #33

### DIFF
--- a/admin/app/view_models/workarea/admin/product_view_model.rb
+++ b/admin/app/view_models/workarea/admin/product_view_model.rb
@@ -33,7 +33,8 @@ module Workarea
       end
 
       def default_category
-        @default_category ||= if categorization.default_model.present?
+        return @default_category if defined? @default_category
+        @default_category = if categorization.default_model.present?
           CategoryViewModel.wrap(categorization.default_model)
         end
       end
--- a/admin/app/view_models/workarea/admin/product_view_model.rb
+++ b/admin/app/view_models/workarea/admin/product_view_model.rb
@@ -33,7 +33,7 @@ module Workarea
       end
 
       def default_category
-        if categorization.default_model.present?
+        @default_category ||= if categorization.default_model.present?
           CategoryViewModel.wrap(categorization.default_model)
         end
       end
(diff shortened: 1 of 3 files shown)


admin/app/view_models/workarea/admin/product_view_model.rb
```diff
@@ -33,7 +33,9 @@ def featured_categories
       end
 
       def default_category
-        CategoryViewModel.wrap(categorization.default_model)
+        @default_category ||= if categorization.default_model.present?
```
bencrouse: This style of memoizing won't work for `nil` since it evals to `false`. You'll want to use the `defined?` style.

admin/test/view_models/workarea/admin/product_view_model_test.rb
```diff
@@ -147,6 +147,14 @@ def test_ignore_inventory
         @view_model = ProductViewModel.new(@product)
         assert(@view_model.ignore_inventory?)
       end
+
+      def test_blank_data
```
bencrouse: Can you add a more specific name here? Maybe `test_blank_categories`

PR #57, opened 2019-09-13
Fix showing comments without authors in admin

eric-pigeon:
Comments generated in plugins don't have an author; update the view to
handle rendering when the `author_id` is nil.


bencrouse:
Whoops, if you want this fixed in v3.4, we'll need another PR to `v3.4-stable`

PR #67, opened 2019-09-16
Add GitHub ribbon to developer docs

### LINKED ISSUE #40: Add GitHub links on developer.workarea.com

meowsus:
Closes #40

No changelog

### DIFF
--- a/docs/source/index.html.erb
+++ b/docs/source/index.html.erb
@@ -1,3 +1,5 @@
+<% @hide_github_badge = true %>
+
 <div class='section [ white-text align-center ]'>
   <h1 class='heading heading--4'>How can we help?</h1>
   <% form_tag '/search.html', class: 'search', method: 'get' do %>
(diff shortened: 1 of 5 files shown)


meowsus:
Screenshot:

Screenshot from 2019-09-16 11-55-01

tubbo:
I've never seen the ribbon down in the bottom right corner, interesting way to solve the problem of "workarea.com" being in the place where ribbons usually go

bencrouse:
Nice! @meowsus can you add this to the homepage too to clean up the visuals on that resources section?
tmp

docs/source/index.html.erb
```diff
@@ -1,3 +1,5 @@
+<% @hide_github_badge = true %>
```
bencrouse: (minor) Try not to make readers think in double negatives (e.g. `@hide_github_badge = false`). Easier to grok if it was `show_github_badge = true`).

PR #69, opened 2019-09-17
Fix Typo in Release Notes Index

tubbo:
Fixes a small typo in the release notes index


tubbo:
if we need the milestone i'll add it back, but since docs are going to be built from github actions going forward, i don't think docs updates will need to be in a release or anything.

PR #76, opened 2019-09-20
Add order confirmation append point

jyucis:
Adds additional append point to the order confirmation page view underneath default order confirmation text.


storefront/app/views/workarea/storefront/checkouts/confirmation.html.haml
```diff
@@ -8,6 +8,7 @@
     = render_content_blocks(@content.content_blocks_for('confirmation'))
 
   %p= t('workarea.storefront.checkouts.confirmation_text', email: @order.email)
+  = append_partials('storefront.order_additional_information')
```
bencrouse: renaming to refer to its location instead of possible use-case
```suggestion
  = append_partials('storefront.checkout_confirmation_text')
```

PR #110, opened 2019-09-25
Fix Self-Referential Category Rules

### LINKED ISSUE #52: Fix combining category and not in category product rules
**Describe the bug**
When you create product rules with a category rule for multiple categories and another rule to exclude a category, it doesn't return the correct products.

**To Reproduce**
See attached test

**Expected behavior**
See attached test

 - Workarea Version: `master`

tubbo:
Adding the same ID to a category product rule matching the product list
that contains it results in some wonky results coming back. This was
originally diagnosed as an issue when combining category rules, but in
reality, it has to do with an admin mis-using the product rules
interface and perhaps accidentally using the category's own ID in a
product rule. To prevent this from happening, clean the
`product_list.id` from the value if a category rule is created or
updated.

Fixes #52

### DIFF
--- a/admin/app/views/workarea/admin/product_rules/fields/_category.html.haml
+++ b/admin/app/views/workarea/admin/product_rules/fields/_category.html.haml
@@ -2,4 +2,4 @@
 Name
 = select_tag 'product_rule[operator]', options_for_select([['equal', 'equal'], ['not equal', 'not_equal']], rule.operator)
 - categories = Workarea::Catalog::Category.any_in(id: rule.terms)
-= select_tag 'product_rule[value]', options_from_collection_for_select(categories, 'id', 'name', rule.terms), multiple: true, data: { remote_select: { source: catalog_categories_path, options: { placeholder: 'Choose a category...' } }.to_json }
+= select_tag 'product_rule[value]', options_from_collection_for_select(categories, 'id', 'name', rule.terms), multiple: true, data: { remote_select: { source: catalog_categories_path(exclude_ids: rule.product_list.id), options: { placeholder: t('workarea.admin.product_rules.fields.category.choose') } }.to_json }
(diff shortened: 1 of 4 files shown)


tubbo:
after talking with @ryaan-anthony at mechanics last night, we determined that the real root cause of the issue was an ID in the product rule that matched the ID of the category. combining category rules that feature categories with different IDs than the parent works fine, it's just when one of those IDs matches the parent that this subtle bug is uncovered.

admin/app/views/workarea/admin/product_rules/fields/_category.html.haml
```diff
@@ -2,4 +2,5 @@
 Name
 = select_tag 'product_rule[operator]', options_for_select([['equal', 'equal'], ['not equal', 'not_equal']], rule.operator)
 - categories = Workarea::Catalog::Category.any_in(id: rule.terms)
-= select_tag 'product_rule[value]', options_from_collection_for_select(categories, 'id', 'name', rule.terms), multiple: true, data: { remote_select: { source: catalog_categories_path, options: { placeholder: 'Choose a category...' } }.to_json }
+- current = Workarea::Search::Admin.for(rule.product_list)
```
bencrouse: if you did the part to get the ES document ID in the controller, you could write a test for it!

admin/app/views/workarea/admin/product_rules/fields/_category.html.haml
```diff
@@ -2,4 +2,5 @@
...
+= select_tag 'product_rule[value]', options_from_collection_for_select(categories, 'id', 'name', rule.terms), multiple: true, data: { remote_select: { source: catalog_categories_path(exclude_ids: current.id), options: { placeholder: 'Choose a category...' } }.to_json }
```
mttdffy: i18n that placeholder

bencrouse:
Thanks for fixing my style @tubbo

PR #113, opened 2019-09-25
Fix Product URL In Breadcrumbs

### LINKED ISSUE #83: Fix product_url in breadcrumb Schema.org structured data
@sperks38 reported a bug around `product_urls` in the breadcrumb structured data. Triage and fix.

tubbo:
The `storefront_url_for` method doesn't handle models other than taxons,
but the Schema.org helpers use it to render breadcrumb URLs in the
`BreadcrumbList` for any model that's in the breadcrumbs. To prevent
incorrect URLs from showing up in the breadcrumbs, the
`Navigation::Breadcrumbs` class has been modified to accept a model
object as its `:last` parameter, instead of just a name, to be added to
an arbitrary `Navigation::Taxon` created for the purpose of rendering
both the name and URL of the final navigation taxon. This wasn't needed
prior to the introduction of Schema.org's `BreadcrumbList`, because the
final URL of breadcrumbs was always left out. The helper methods that
render the breadcrumbs will continue to leave out the final taxon's URL,
but for breadcrumbs in Schema.org, the URL will now be included.

(#83)

### DIFF
--- a/core/app/models/workarea/navigation/breadcrumbs.rb
+++ b/core/app/models/workarea/navigation/breadcrumbs.rb
@@ -13,7 +13,16 @@ module Workarea
 
       def initialize(navigable, last: nil)
         @navigable = navigable
-        @last = Taxon.new(name: last) if last.present?
+        @last = if last.class.include?(Navigable)
+                  Taxon.new(
+                    name: last.name,
+                    navigable: last,
+                    navigable_slug: last.slug,
+                    parent: breadcrumb_taxons.last
+                  )
+                elsif last.present?
+                  Taxon.new(name: last, parent: breadcrumb_taxons.last)
+                end
       end
 
       def to_global_id
--- a/core/test/models/workarea/navigation/breadcrumbs_test.rb
+++ b/core/test/models/workarea/navigation/breadcrumbs_test.rb
@@ -23,10 +23,20 @@ module Workarea
       end
 
       def test_last
-        breadcrumbs = Breadcrumbs.new(@navigable, last: 'Test Taxon')
+        model = create_product
+        breadcrumbs = Breadcrumbs.new(@navigable, last: model)
+
         assert(breadcrumbs.last.is_a?(Navigation::Taxon))
         assert_equal(4, breadcrumbs.length) # including home
-        assert_equal('Test Taxon', breadcrumbs.last.name)
+        assert_equal(model, breadcrumbs.last.navigable)
+        assert_equal(model.name, breadcrumbs.last.name)
+
+        model = create_inventory
+        breadcrumbs = Breadcrumbs.new(@navigable, last: model.id)
+
+        assert(breadcrumbs.last.is_a?(Navigation::Taxon))
+        assert_equal(4, breadcrumbs.length) # including home
+        assert_equal(model.id, breadcrumbs.last.name)
       end
 
       def test_selected
--- a/storefront/app/view_models/workarea/storefront/product_view_model.rb
+++ b/storefront/app/view_models/workarea/storefront/product_view_model.rb
@@ -1,7 +1,6 @@
 module Workarea
   module Storefront
     class ProductViewModel < ApplicationViewModel
-
       def self.wrap(model, options = {})
         if model.is_a?(Enumerable)
           model.map { |m| wrap(m, options) }
@@ -18,12 +17,12 @@ module Workarea
           if options[:via].present?
             Navigation::Breadcrumbs.from_global_id(
               options[:via],
-              last: model.name
+              last: model
             )
           else
             Navigation::Breadcrumbs.new(
               default_category,
-              last: model.name
+              last: model
             )
           end
       end
--- a/storefront/test/helpers/workarea/storefront/schema_org_helper_test.rb
+++ b/storefront/test/helpers/workarea/storefront/schema_org_helper_test.rb
@@ -8,14 +8,11 @@ module Workarea
 
       def test_breadcrumb_list_schema
         product = create_product
-        # so the product will have a parent.
-        create_category(product_ids: [product.id.to_s]).tap do |cat|
-          create_taxon(navigable: cat)
-        end
+        category = create_category(product_ids: [product.id]) # so the product will have a parent
+        create_taxon(navigable: category)
+
         view_model = ProductViewModel.wrap(product)
-        breadcrumbs = view_model.breadcrumbs.map do |taxon|
-          [taxon.name, storefront_url_for(taxon)]
-        end
+        breadcrumbs = view_model.breadcrumbs.map { |t| [t.name, storefront_url_for(t)] }
         schema = breadcrumb_list_schema(breadcrumbs)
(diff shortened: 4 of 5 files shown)


storefront/app/helpers/workarea/storefront/navigation_helper.rb
```diff
@@ -65,22 +65,25 @@ def link_to_menu(menu)
         end
       end
 
-      def storefront_path_for(taxon)
-        if taxon.url?
-          taxon.url
-        elsif taxon.root?
+      def storefront_path_for(model)
```
bencrouse: This wasn't intended to allow passing any model to it. We should correct it upstream. Where are we calling this without a taxon?

storefront/test/helpers/workarea/storefront/schema_org_helper_test.rb
```diff
@@ -0,0 +1,24 @@
...
+        page = create_page
+        url_taxon = create_taxon(url: 'https://www.example.com')
+        navigable_taxon = create_taxon(navigable: page)
+        inventory = create_inventory
+
+        assert_equal(product_url(product, host: Workarea.config.host), breadcrumb_url_for(product))
+        assert_equal(page_url(page, host: Workarea.config.host), breadcrumb_url_for(page))
+        assert_equal('https://www.example.com', breadcrumb_url_for(url_taxon))
+        assert_equal(page_url(page, host: Workarea.config.host), breadcrumb_url_for(navigable_taxon))
+        assert_raises(NoMethodError) { breadcrumb_url_for(inventory) }
```
tubbo: I don't know if we should warn people or render a better error here...the NoMethodError produced when you add something to breadcrumbs that isn't routed at the top level doesn't really tell you much about why it happened...

storefront/app/views/layouts/workarea/storefront/application.html.haml
```diff
@@ -102,7 +102,7 @@
             .breadcrumbs{ role: 'navigation' }
               = yield :breadcrumbs
           - elsif @breadcrumbs.present?
-            = render_schema_org(breadcrumb_list_schema(@breadcrumbs.map { |t| [t.name, storefront_url_for(t)] }))
+            = render_schema_org(breadcrumb_list_schema(@breadcrumbs.map { |t| [t.name, breadcrumb_url_for(t)] }))
```
bencrouse: It feels like the problem should be handled more gracefully upstream so we don't have to do all the string munging to figure out the route. Where is the bad data coming from?

storefront/test/helpers/workarea/storefront/schema_org_helper_test.rb
bencrouse: I agree, I think we should look at what the data is upstream causing this situation and see if we can change that.

core/app/models/workarea/navigation/breadcrumbs.rb
```diff
@@ -13,7 +13,7 @@ def self.from_global_id(via, last: nil)
 
       def initialize(navigable, last: nil)
         @navigable = navigable
-        @last = Taxon.new(name: last) if last.present?
+        @last = Taxon.new(name: last.name, navigable: last) if last.present?
```
bencrouse: Can we have this gracefully degrade so keep the same API? Then we don't need to change the `ProductViewModel`.
```suggestion
        @last = if last.try(:navigable?)
          Taxon.new(name: last.name, navigable: last)
        elsif last.present?
          Taxon.new(name: last)
        end
```

core/test/models/workarea/navigation/breadcrumbs_test.rb
```diff
@@ -23,10 +23,12 @@ def test_collection
       end
 
       def test_last
```
bencrouse: After the suggestion above, you should be able to reduce this to just testing a navigable as `last`.

storefront/app/view_models/workarea/storefront/product_view_model.rb
```diff
@@ -1,7 +1,6 @@
 module Workarea
   module Storefront
     class ProductViewModel < ApplicationViewModel
-
```
bencrouse: Can remove changes to this file after suggestion above.

bencrouse:
This is coming along well!

storefront/app/view_models/workarea/storefront/product_view_model.rb
tubbo: that would kinda defeat the purpose of all this, because a string isn't `.navigable?`, so we still need to pass the full product model into the breadcrumbs as `:last` in order to generate the right URL.

core/app/models/workarea/navigation/breadcrumbs.rb
tubbo: I guess I'm OK with this but again this defeats the purpose of the fix to schema.org helpers, which is what this ticket was about in the first place. Definitely would help with backwards-compatibility in a patch, though, so I think this is OK...but I can't undo the changes to ProductViewModel as a result since it needs to pass in the full model so the final Taxon knows what URL to generate in `storefront_path_for`.

core/app/models/workarea/navigation/breadcrumbs.rb
```diff
@@ -13,7 +13,11 @@ def self.from_global_id(via, last: nil)
 
       def initialize(navigable, last: nil)
         @navigable = navigable
-        @last = Taxon.new(name: last) if last.present?
+        @last = if last.try(:navigable?)
```
bencrouse: Judging from the build, I think I forgot we'd need a `def navigable?` in the `Navigable` module.

storefront/app/view_models/workarea/storefront/product_view_model.rb
bencrouse: Do the tests pass without the change?

tubbo: fixed it by checking whether `last.class.includes? Navigable` so we don't have to write an additional method.

storefront/app/view_models/workarea/storefront/product_view_model.rb
```diff
@@ -18,12 +17,12 @@ def breadcrumbs
           if options[:via].present?
             Navigation::Breadcrumbs.from_global_id(
               options[:via],
-              last: model.name
+              last: model
```
bencrouse: So why do this unnecessary change? Better to reduce churn for upgrades if this doesn't affect anything.

tubbo: this isn't unnecessary, it's what fixes the schema

tubbo: again, this isn't unnecessary. i wrote in an above conversation:

> that would kinda defeat the purpose of all this, because a string isn't `.navigable?`, so we still need to pass the full product model into the breadcrumbs as `:last` in order to generate the right URL.

basically, i need `:last` to be a model in order for the Schema.org helper methods to render the right URL for the last Taxon.

Issue #116, opened 2019-09-25
Inverse aspect ratio not being stored on Dragonfly models


bencrouse:
**Describe the bug**
`product_image.inverse_aspect_ratio` isn't being saved in MongoDB as a magic Dragonfly attribute. This is causing requests to fetch the asset from S3 on every request.

**Expected behavior**
The value should be stored in a field on the model, like other magic Dragonfly attributes.

**Workarea Setup (please complete the following information):**
 - Workarea Version: v3.4.16

bencrouse:
@tubbo this one is top priority!

PR #118, opened 2019-09-26
Add Inverse Aspect Ratio To Product Image Fields

tubbo:
Populate the `:image_inverse_aspect_ratio` automatically using
Dragonfly, in order to reduce the amount of requests made to S3 in order
to find out this information. This way, Dragonfly can store more assets
in the cache.

(#116)


tubbo:
> Can you do this for `Content::Asset` too?

@bencrouse it's already there. https://github.com/workarea-commerce/workarea/blob/master/core/app/models/workarea/content/asset.rb#L30

PR #119, opened 2019-09-26
Remove usage of undefined #second method on Array

mttdffy:


mttdffy:
Figured we should take docs out of the gem while I'm thinking of it. 

Gem build went from 22.3MB to 173kb. 🎉

workarea.gemspec
```diff
@@ -10,7 +10,7 @@ Gem::Specification.new do |s|
   s.summary     = 'The Workarea Commerce Platform'
   s.description = 'Workarea is an enterprise-grade Ruby on Rails commerce platform.'
 
-  s.files = Dir['README.md', 'CHANGELOG.md', 'docs/**/*', '.rubocop.yml', 'docker-compose.yml']
```
bencrouse: 💯 😄

Issue #95, opened 2019-09-23
Alt text UI refinements


ebigler:
Remove alt text defaults from content DSL
Always show alt text message on content image fields
Add alt text to asset selection

PR #120, opened 2019-09-26
Surface Asset alt text and behavior within Content Blocks (#95)

meowsus:
In an effort to make the recent updates to alt text overridding in
Content Blocks a bit clearer, alt text is now being output:

- On the content assets index view
- In the title for a content asset summary

Default alt text has been removed from the Content block DSL, which
makes the default text come directly from the Asset itself.

The help text displayed on Asset Content Blocks always appears now,
better explaining the behavior of this feature.

### DIFF
--- a/admin/app/views/workarea/admin/content_assets/index.html.haml
+++ b/admin/app/views/workarea/admin/content_assets/index.html.haml
@@ -55,6 +55,7 @@
                 = label_tag 'select_all', t('workarea.admin.bulk_actions.select_all'), class: 'checkbox__label'
             %th= t('workarea.admin.fields.image')
             %th= t('workarea.admin.fields.name')
+            %th= t('workarea.admin.fields.alt_text')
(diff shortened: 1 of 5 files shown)


admin/app/views/workarea/admin/content_assets/index.html.haml
```diff
@@ -77,6 +78,11 @@
                 = upcoming_changesets_icon_for(result)
                 = comments_icon_for(result)
                 = favicon_icon_for(result)
+              %td
+                - if result.image?
+                  = result.alt_text
```
bencrouse: Should we be worried about really long values messing up display here?

meowsus: Yes, let's.

meowsus:
> Seems to be a few build issues...

@duffy hoping to fix this in #129

PR #121, opened 2019-09-26
Remove Refund Tests

tubbo:
Since we're no longer able to regenerate VCR cassettes at-will (due to
credentials needing to be scrubbed before pushing to GitHub), this
configuration setting is no longer necessary, and furthermore, could
potentially prevent legitimate tests from running and catching bugs in
the wild. Remove the configuration and the guard clause in the refund
tests for `Workarea::Payment::CreditCardIntegrationTest`, and fix any
build failures in plugins as they come up.

### DIFF
--- a/core/lib/workarea/configuration.rb
+++ b/core/lib/workarea/configuration.rb
@@ -140,11 +140,6 @@ module Workarea
 
       config.credit_card_issuers['bogus'] = 'Test Card' unless Rails.env.production?
 
-      # Most credit card processors won't let a capture/purchase be refunded until
-      # the transaction settles. (Most processors settle once a day).  If your processor
-      # allows immediate refuding you can include extra tests around refund operations.
-      config.run_credit_card_refund_tests = false
-
       # Determines the order in which discounts are calculated.
       config.discount_application_order = %w(
         Workarea::Pricing::Discount::Product
--- a/core/test/models/workarea/payment/credit_card_integration_test.rb
+++ b/core/test/models/workarea/payment/credit_card_integration_test.rb
@@ -76,58 +76,6 @@ module Workarea
         end
       end
 
-      def test_auth_capture_refund
-        pass && return unless Workarea.config.run_credit_card_refund_tests
-
-        VCR.use_cassette 'credit_card/auth_capture_refund' do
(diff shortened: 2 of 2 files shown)


eric-pigeon:
It's impossible for those tests to pass depending on the credit card gateway being implemented.  If the gateway doesn't allow refunding of unsettled transactions, like braintree or authorize.net cim(or the majority of credit card gateways) these tests will only lead to confusion and become a time sink.  These tests are only capable of passing in rare scenarios.

tubbo:
@eric-pigeon if most payment plugins don't actually take advantage of these tests, then they should be removed from base and added into each individual plugin that does support the functionality.

**edit:** a quick search through all the proprietary plugin code indicates that this config is only set to `true` in two plugins. one of these plugins re-implements the entire test, meaning the config setting is useless, and the other plugin actually uses the config as intended.

I propose that we remove the tests from base entirely, and copy them into the one plugin that actually uses them.

eric-pigeon:
@tubbo I checked all of the Repos I had locally; 4 of them set this config to true.  Cybersource and Checkoutdotcom run the refund tests without modifications.  Payeezy did decorate the one test but the changes look like no functionally was actually changed, I would investigate if they decoration in the plugin is actually needed.  And finally Flowio weirdly copied the exact tests but is able to make use of the refund tests for their faux credit card active merchant gateway.  The intention is to set up developers with the highest chance of success; these tests are the baseline for every credit card integration.  If these tests pay there is a good chance your integration is correct, the only exception would be refunding which might not be automatable which is why the config exists.

bencrouse:
Sounds like these aren't general enough (2/4 plugins that use them can use them as-is), so they aren't super helpful. But we do need to write good coverage in plugins for what they support.

I did an ack in my projects directory, doesn't look like any builds are using it directly.

@tubbo are you going to go through those 4 plugins and update their tests to correspond with this change?

tubbo:
Searching stash only returns 2 repos that have the config set to `true`, workarea-checkoutdotcom and workarea-flowio. In flow, the tests are decorated anyway, so no need to change anything there. I copied the tests from base into checkoutdotcom and removed the config setting. Payeezy decorated the tests to make them run, but didn't actually set the config to true, so I don't think any changes are needed there either.

PR #123, opened 2019-09-26
Fix duplication in README

chriscressman:
No changelog


chriscressman:
I'm guessing this happened during a merge, and I think this is what was intended.

PR #128, opened 2019-09-30
Pass Options To `Storefront::UserActivityViewModel`

tubbo:
This was an oversight that got caught and fixed in the `flow-io` plugin,
but should really be in base since it will allow more control over the
product summaries on the recent views action. The `view_model_options`
were not getting passed into the `UserActivityViewModel`, and thus the
`ProductViewModel` instances that it creates, causing some stale content
to appear in the view.

### DIFF
--- a/storefront/test/integration/workarea/storefront/recent_views_integration_test.rb
+++ b/storefront/test/integration/workarea/storefront/recent_views_integration_test.rb
@@ -3,34 +3,26 @@ require 'test_helper'
 module Workarea
   module Storefront
     class RecentViewsIntegrationTest < Workarea::IntegrationTest
-      include Storefront::IntegrationTest
-
       def test_show
         password = 'W3bl1nc!'
         user = create_user(password: password)
         category = create_category
-        href = storefront.product_path(product, via: category.id)
-
-        post storefront.login_path, params: {
-          email: user.email,
-          password: password
-        }
-
-        assert_redirected_to(storefront.users_account_path)
-        refute(session[:user_id].blank?)
-
+        product = create_product
         Metrics::User.save_affinity(
           id: user.email,
           action: 'viewed',
           product_ids: [product.id]
         )
 
-        get storefront.recent_views_path(via: category.id)
+        post storefront.login_path,
+          params: { email: user.email, password: password }
 
-        assert_response(:success)
-        assert_select('.recent-views__section--products')
-        assert_select('.product-summary')
-        assert_select(%(.product-summary__name a[href="#{href}"]))
+        assert_redirected_to(storefront.users_account_path)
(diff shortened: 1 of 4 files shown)


storefront/test/integration/workarea/storefront/recent_views_integration_test.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        assert_response(:success)
+        assert_select('.recent-views__section--products')
+        assert_select('.product-summary')
+        assert_select(%(.product-summary__name a[href="#{href}"]))
+      end
+
+      private
+
+      def sign_in(user, password = 'W3bl1nc!')
```
bencrouse: You'll want to specify this password to `create_user` in case something changes. Also not sure a separate method is really adding much here given this is only used once.

storefront/test/integration/workarea/storefront/recent_views_integration_test.rb
```diff
@@ -0,0 +1,40 @@
...
+        category = create_category
+        href = storefront.product_path(product, via: category.id)
+
+        sign_in(user)
+        Metrics::User.save_affinity(
+          id: user.email,
+          action: 'viewed',
+          product_ids: [product.id]
+        )
+        get storefront.recent_views_path(via: category.id)
```
bencrouse: Trying to understand, where does the storefront do this? I'm blanking on it.

storefront/test/view_models/workarea/storefront/user_activity_view_model_test.rb
```diff
@@ -19,7 +18,10 @@ def reset_display_count
       def test_products
         product_one = create_product
         product_two = create_product
-        product_three = create_product
+        product_three = create_product(
```
bencrouse: What do these have to do with the change? I'm not following...

storefront/test/integration/workarea/storefront/recent_views_integration_test.rb
tubbo: it doesn't, but flow does.

storefront/test/view_models/workarea/storefront/user_activity_view_model_test.rb
tubbo: differentiates this product from the other two

storefront/test/integration/workarea/storefront/recent_views_integration_test.rb
tubbo: to clarify, flow doesn't actually send this param AFAIK, but it might send other params that affect how a product is priced, and `:via` is the only param that I saw that I could actually test for in the view.

PR #131, opened 2019-09-30
Check session for user_id over cookies

mttdffy:
The RoutingConstraints::SuperAdmin protects access to the Sidekiq
web interface. The constraint needed to be updated to reflect the
storing of user_id in session as of v3.5


mttdffy:
Closed this thinking it wouldn't work. Tested it. It did.

PR #132, opened 2019-09-30
[WIP] Heroku Review Apps

tubbo:
Heroku has a really neat feature called **review apps** that will
automatically deploy a "disposable" application whenever a new pull
request is opened on Workarea. Review apps can help us in our workflow
to reduce some of the bottlenecks that occur around release time, namely
when a bunch of merged pull requests need to be QAed all at once, or the
qa/edge staging servers aren't behaving appropriately. With review apps,
anyone on the team can QA an open pull request in an isolated,
disposable Workarea demo application with fresh seeds, and even include
QA notes in code review rather than several days after the changes have
been merged into master and our stable branches. QA will no longer be
blocked by a mergedown from v3.4-stable to master, because a review app
will already be available for that pull request until it's merged in.
Additionally, with review apps and Heroku's free tier, we will no longer
need to manage our own v3-qa and v3-edge servers, which reduces costs
for Workarea as a company.

## In The Wild

Recently, I got review apps working for a personal project because it was already running on Heroku. You can see how they've helped in this open pull request: https://github.com/waxpoetic/brotherly/pull/90

## Possible Downsides

- Work that is interdependent on other branches being merged will not be
  able to be tested in this manner, but review apps won't block the normal
  code review flow, so we'll still be able to merge pull requests without
  a successful deployment. This is just something we'll have to work out
  on the shared QA server (which is what the review apps are based off)
  when everything is merged.
- When working on issues that stem from the interaction between multiple
  plugins, you'll need to add the plugin to the app's Gemfile in
  Dockerfile so it gets installed along with the plugin you've made code
  changes in. This doesn't happen all that often, but it is one more
  necessary step in order to see your changes on the review app.

## TODO

- [ ] I need admin access to this repository in order to enable review
      apps, even though Heroku has been authorized for this organization
      under my name. Before I went around asking about this, I figured I
      would write something up and show you guys what this is all about.
- [ ] Since Heroku doesn't provide an Elasticsearch service that is
      compatible with our platform, we'll need to host an authenticated
      Elasticsearch cluster that these Heroku apps can connect to. The URL
      won't be public, and it will be behind HTTP Basic Auth so no one else
      can connect to it. When Heroku apps are destroyed after a
      successful merge (or no deploys occurring within a given time
      frame), the indexes created during its lifespan will be deleted so
      we won't have to continually manage the server.
- [ ] A `Dockerfile` for plugins has not yet been written, but my plan
      is to make it very similar to the one in base, except it will
      just install the latest `workarea` gem and load the plugin from a
      directory on the image called `/plugin`. Each plugin will have its
      own separate Heroku app and pipeline, but use the same
      Elasticsearch server as described above.

## What's Changed

- **Add a `Dockerfile` to the root for building the application on Heroku.**
  This is very similar to `demo/Dockerfile`, but instead of installing
  the latest released version of Workarea, it copies the current state
  of the repo to `/workarea` and loads gems from there.
- **Configure Mongo and Redis with environment variables that end in
  `_URL`.** By default, Heroku addons provide `_URL` environment variables
  to access the various services that you've provisioned, this just cuts
  down on boilerplate in the Dockerfile and makes it easier to launch
  Workarea on Heroku.
- **Add `app.json` and `heroku.yml` for configuring the application on
  Heroku automatically.** These files tell Heroku to build from the
  `Dockerfile` mentioned above, as well as provide the required
  configuration needed for deploying the app to Heroku.
- **Add "Deploy to Heroku" button in case anyone else wants to deploy the
  current state of Workarea to their Heroku instance.** ~~I wish there was
  some way of differentiating our `app.json` used for review apps and
  one that would always load the demo image, but I'm not sure how to do
  that. This button is ancillary to the rest of the changes, but it's a
  nice thing to have now that we have the capability.~~ This is now pointing to `demo/` on v3.4-stable
- Heroku review apps have arbitrary domain names, so when generating a
  new app in a Heroku environment, **automatically configure the domain
  name to `https://${HEROKU_APP_NAME}.herokuapp.com`**, which is where the
  app will actually be served from. `$HEROKU_APP_NAME` will also be what
  `Workarea.config.site_name` is set to, in order to differentiate it in
  the shared Elasticsearch server from other applications.


app.json
```diff
@@ -0,0 +1,46 @@
...
+      "plan": "heroku-redis:hobby-dev",
+      "as": ["WORKAREA_REDIS", "WORKAREA_REDIS_CACHE"]
+    }
+  ],
+  "buildpacks": [
+    {
+      "url": "https://github.com/heroku/heroku-buildpack-ruby"
+    },
+  ],
+  "env": {
```
tubbo: in review apps, the settings described here will be provided by the parent app that is already running. when deploying to heroku, a user will have to fill in this information before they can deploy (except for `HEROKU_APP_NAME` and `HEROKU_PARENT_APP_NAME`)

app.json
```diff
@@ -0,0 +1,47 @@
...
+      "required": true
+    },
+    "HEROKU_APP_NAME": {
+      "required": "true"
+    },
+    "HEROKU_PARENT_APP_NAME": {
+      "required": "true"
+    }
+  },
+  "scripts": {
```
tubbo: the following two scripts set up the DB when created and remove Elasticsearch indexes when destroyed. this is to reduce load on the shared ES server that all review apps will be connected to.

core/lib/generators/workarea/install/templates/initializer.rb.erb
```diff
@@ -5,7 +5,11 @@ Workarea.configure do |config|
   config.host = {
     'test' => 'www.example.com',
     'development' => 'localhost:3000',
-    'production' => 'www.<%= app_name.dasherize %>.com' # TODO
+    'production' => if ENV['HEROKU_APP_NAME'].present?
```
tubbo: I'm definitely not happy about adding more code to the initializer, but I honestly couldn't see any other way to do it. Open to suggestions on a better way of setting the site host automatically when we're on Heroku

bencrouse:
In general this could be cool, but I'm not really interesting in making all of these invasive changes to support it right now (there hasn't been any demand for Heroku so far). I think options forward include:

* stripping this back to minimal changes
* finding a way to do it within one of our K8s clusters
* create a separate repo to support this instead of cluttering up base

and I think it still leaves some significant process changes in question:

* how do we handle open source plugins?
* how do we handle commerce cloud plugins?
* how do we handle different themes?

If this is a direction we feel we should go, we should chat about options and think it over as a team. If you're trying to make progress right now, I'd suggest creating a separate repo as a proof of concept.

tubbo:
Yup, I wasn't intending on this to be a "final version" of the idea, but I thought I'd show you folks where I'm at and we can discuss whether we want to do this, and if so, how. I responded to some of your concerns below...

> how do we handle plugins?

I laid out how we would handle plugins in the PR description, and themes kinda go with that territory. Each plugin and theme gets their own review apps pipeline, and a Dockerfile, which loads the `workarea/demo` Docker image and adds the current gem code into `/plugin` on the container. This actually would let us QA different themes simultaneously, rather than having to manually install a theme on a QA server, or have a QA server running all the time that we only use every once in a while. I don't see a need to differentiate between commerce cloud and open source plugins, because right now we don't do that and our QA servers are entirely open to the public. One would have to know the exact PR number to view a QA app for a repo that they don't have access to. That said, pulling in basic auth to protect the QA server(s) isn't out of the question, just didn't feel like it would be necessary since we don't authenticate users on v3-edge today.

> stripping this back to minimal changes

With the exception of the "Deploy to Heroku" button, I tried to be as non-invasive as I could and make only the changes that seemed necessary to get review apps going. What do you see as non-essential in these changes?

> finding a way to do it within one of our K8s clusters

You mean re-implement Heroku's Review Apps system? Or something else? There's a lot of work to be done if we want to reinvent the wheel and build our own thing to do review apps, and I'm not sure what the benefit would be...

> create a separate repo to support this instead of cluttering up base

I thought about that at first. It breaks down when we're talking about different plugins, unless this separate repo took the place of our megabuild or something, and would load every plugin onto it. But the flaw in that is if you're trying to QA certain plugins, like payment integrations or themes, where there can only be one of them in the Gemfile at any time...It would be really sweet if we didn't have to manually manage our QA servers at all, each repo came with its own on-demand QA apps that would be triggered when we made a PR.

bencrouse:
I'm closing this until we want to explore this option again.

Issue #102, opened 2019-09-24
OOTB Segments


ebigler:
Segments functionality will be improved and expanded. Several life cycle segments (for example 'first time visitor') will exist out of the box and collect analytics data per segment. Segment controls will also be added to releases.

PR #135, opened 2019-10-01
Allow setting active by segment

bencrouse:
This allows configuring releasable resources to be active only for
certain segments. If no segments are specified, it will be active
globally. If segments are specified, only those segments will be able to
see it.

For #102


admin/app/assets/javascripts/workarea/admin/modules/select2.js
```diff
@@ -0,0 +1,30 @@
+/**
```
mttdffy: im confused by the addition of this file, but I'll leave the judging to Curt

admin/app/assets/javascripts/workarea/admin/modules/toggle.js
```diff
@@ -0,0 +1,21 @@
+/**
```
mttdffy: thought we have this behavior already, but again I'll defer to Curt

admin/app/controllers/workarea/admin/segments_controller.rb
```diff
@@ -3,11 +3,12 @@ module Admin
     class SegmentsController < Admin::ApplicationController
       required_permissions :people
 
-      before_action :find_segment, except: :index
+      before_action :find_segment, except: %i(index move)
```
mttdffy: i don't see a move action?

admin/config/routes.rb
```diff
@@ -362,10 +362,15 @@
     resources :segments, except: [:new, :create] do
       resources :rules, except: [:show, :new, :edit], controller: 'segment_rules'
 
+      collection do
+        post :move
```
mttdffy: no matching action?

core/app/models/workarea/content/block_draft.rb
```diff
@@ -29,7 +32,9 @@ def typecast_data!
       end
 
       def to_block
-        content.blocks.build(attributes.except('content_id'))
+        result = Content::Block.instantiate(as_document.except('content_id'))
```
mttdffy: TIL `.instantiate` is a thing. Whats the significance over `.new`?

storefront/app/view_models/workarea/storefront/recommendations_view_model.rb
```diff
@@ -34,6 +32,12 @@ def product_ids
       def result_count
         raise NotImplementedError
       end
+
+      private
+
+      def find_displayable_products(ids)
+        Catalog::Product.find_ordered(ids).select(&:active?).select(&:purchasable)
```
mttdffy: should use `purchasable?` as that factors in variant activeness, too.

admin/app/assets/javascripts/workarea/admin/modules/toggle.js
bencrouse: yeah, I thought so too 😞

admin/app/assets/javascripts/workarea/admin/modules/select2.js
bencrouse: The idea is to allow you to make a pure instance of select2 for things like inline options like `allowClear`, etc. `remoteSelects` is written specifically for remote stuff: https://github.com/workarea-commerce/workarea/blob/master/admin/app/assets/javascripts/workarea/admin/modules/remote_selects.js

admin/app/controllers/workarea/admin/segments_controller.rb
bencrouse: Yep, cruft from when segments had priority.

admin/config/routes.rb
bencrouse: Yep, cruft.

core/app/models/workarea/content/block_draft.rb
bencrouse: Yeah, this was a TIL I learned for me too. Basically like `new`, but assumes the attributes are raw, as-if from the DB.

bencrouse: This is important here because `active_segment_ids` are localized.

bencrouse:
@meowsus can you take a look? I'd like to get this merged ASAP so QA can begin.

admin/app/assets/javascripts/workarea/admin/modules/select2.js
meowsus: @mttdffy this is something we had originally, long, long ago, when Select2 was introduced into the Admin, but we had no real use case for it (until now) so it fell by the wayside. As part of #142 I'll better align remoteSelects & select2's APIs so this makes a bit more sense.

admin/app/assets/javascripts/workarea/admin/modules/toggle.js
meowsus: I did as well. I'll triple check as part of #142.

admin/app/views/workarea/admin/segments/index.html.haml
```diff
@@ -6,37 +6,15 @@
       .grid__cell.grid__cell--50
         .view__heading
           = link_to "↑ #{t('workarea.admin.segments.index.dashboard_link')}", people_dashboards_path, class: 'view__dashboard-button'
-          %h1= t('workarea.admin.segments.index.page_title')
+          %h1.heading--no-margin= t('workarea.admin.segments.index.page_title')
```
meowsus: Super minor: `.heading.heading--no-margin`

core/lib/workarea.rb
```diff
@@ -182,6 +182,14 @@ def self.define_content_block_types(&block)
     require_dependency Workarea::Core::Engine.root.join('app', 'models', 'workarea', 'content', 'block_type_definition')
     Content::BlockTypeDefinition.new.instance_eval(&block)
   end
+
+  # Use this deprecation to warn about the next minor release.
+  #
+  # @return [ActiveSupport::Deprecation]
+  #
+  def self.deprecation
+    @deprecation ||= ActiveSupport::Deprecation.new('3.6', 'Workarea')
```
meowsus: I feel like passing the version to the `deprecation` method would help us grok for deprecations more easily, much like we do for `TODO v3.6` 

```ruby
def self.deprecation(version)
  @deprecation ||= ActiveSupport::Deprecation.new(version, 'Workarea')
end
```

That way we're not always forgetting  to bump this version in this method every minor, and we can more easily `ack "deprecation('3.6')"` at the same time we `ack 'TODO 3.6'`.

Just a thought.

meowsus: Loving this standardization tho. I always find myself looking up how we do this when I need to deprecate stuff.

PR #136, opened 2019-10-01
Fix test failure due to iframe loading

bencrouse:
This test has started failing due to Capybara or Selenium not finding
the release select in the iframe. A simple sleep fixes the problem, we
weren't able to track down a proper cause.

We'll be refactoring the admin toolbar away from an iframe in v3.6, so
this will be a temporary hack to fix. We'll remove this at that point in
time.


bencrouse:
@meowsus I didn't know what project to put this on, so I left blank for now.

meowsus:
> @meowsus I didn't know what project to put this on, so I left blank for now.

@bencrouse we abandoned projects when we went to ZenHub.

PR #138, opened 2019-10-02
Add additional segment rules

bencrouse:
For whether the visitor is logged in and whether they're tagged.

No changelog

This is pointing at my branch for activeness for now.


mttdffy:
Other than the build this looks good to me

Issue #91, opened 2019-09-23
Handle taxes for digital items


ebigler:

ebigler:
pass the billing address instead of the shipping address for calculating taxes for digital products. There is also a discourse post about this topic.

PR #139, opened 2019-10-02
Apply tax to items that do not require shipping

mttdffy:
* Adds Payment lookup to pricing request
* Modifies TaxCalculator to check shipped and not shipped items
* Renames TaxApplier to ShippedTaxApplier, Uses TaxApplier for not shipped items

Work for #91

### DIFF
--- a/core/app/models/workarea/pricing/calculators/tax_calculator.rb
+++ b/core/app/models/workarea/pricing/calculators/tax_calculator.rb
@@ -5,6 +5,11 @@ module Workarea
         include Calculator
 
         def adjust
+          adjust_shipped_items_tax
+          adjust_not_shipped_items_tax
+        end
+
+        def adjust_shipped_items_tax
           shippings.each do |tmp_shipping|
             next unless tmp_shipping.address.present?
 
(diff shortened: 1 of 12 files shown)


core/app/models/workarea/pricing/calculators/tax_calculator.rb
```diff
@@ -5,23 +5,50 @@ class TaxCalculator
...
+          adjust_not_shipped_items_tax
+        end
+
+        def adjust_shipped_items_tax
           shippings.each do |tmp_shipping|
             next unless tmp_shipping.address.present?
 
             adjustments_to_tax = price_adjustments_for(tmp_shipping)
-            TaxApplier.new(tmp_shipping, adjustments_to_tax).apply
+            ShippedTaxApplier.new(tmp_shipping, adjustments_to_tax).apply
```
bencrouse: So this will break all current tax integrations? Why not make the digital items the new case so integrations can gracefully upgrade to include them?

core/app/models/workarea/pricing/shipped_tax_applier.rb
```diff
@@ -0,0 +1,100 @@
...
+        order_item = adjustment._parent
+        partial_shipping_quantity = @shipping.quantities[order_item.id.to_s].to_i
+        return 0 if @shipping.partial? && partial_shipping_quantity.zero?
+
+        total = @adjustments_by_parent[order_item].taxable_share_for(adjustment)
+        total *= partial_shipping_quantity / order_item.quantity.to_f if @shipping.partial?
+        total
+      end
+
+      def calculate_tax_amounts(taxable_amount, rate)
```
bencrouse: This doesn't look needed since you've included `TaxApplication`

core/app/models/workarea/pricing/shipped_tax_applier.rb
```diff
@@ -0,0 +1,100 @@
...
+        {
+          'country_amount' => calculate_tax_amount(taxable_amount, rate.country_percentage),
+          'region_amount' => calculate_tax_amount(taxable_amount, rate.region_percentage),
+          'postal_code_amount' => calculate_tax_amount(taxable_amount, rate.postal_code_percentage)
+        }
+      end
+
+      def calculate_tax_amount(amount, percentage)
```
bencrouse: Same here...

core/test/models/workarea/pricing/calculators/tax_calculator_test.rb
```diff
@@ -112,6 +112,46 @@ def test_assign_shipping_tax_when_tax_does_not_charge_on_shipping
...
+                    data: { 'tax_code' => '001' }
+                  }
+                ]
+              }
+            ]
+          )
+
+          payment = create_payment(
+            id: order.id,
+            address: factory_defaults_config.billing_address
```
bencrouse: It'd be nice to expose this more clearly because we've needed it in so many places. Maybe like `def address_attributes` or something similar?

core/app/models/workarea/pricing/calculators/tax_calculator.rb
mttdffy: have to stop thinking about it as digital items, rather than "not shipped" items. I see your point though.

bencrouse: Yeah, I'm not loving the language around this so far, but I'm not sure what a better word is. What kinds of products aren't shipped and aren't digital?

core/app/models/workarea/pricing/item_tax_applier.rb
```diff
@@ -0,0 +1,41 @@
+module Workarea
+  module Pricing
+    class ItemTaxApplier
```
bencrouse: Nice, this is a good refactor.

PR #141, opened 2019-10-02
Use deprecator

bencrouse:
This'll be easier to digest after #135 merges, might want to wait until then to peep.


bencrouse:
@mttdffy only because I added `Workarea.deprecation` there for some segment-related changes.

Issue #144, opened 2019-10-03
`_id` Omitted from Customized Fields


tubbo:
**Describe the bug**
Any customized fields ending in `_id` will have the `_id` omitted from their instance variable name, which causes issues if that name conflicts with an ivar or method that already exists.

**To Reproduce**
1. Add a customized field called `:product_id`
2. Create your customizations with `MyCustomizations.new('1234', 'product_id' => '5678')`
3. Call `customizations.product_id`, it will be "1234".

**Expected behavior**
Since the `product_id` param overrides the `@product_id` assigned by the first argument, the value of `#product_id` in the customization object should be "5678". Digging in reveals that the value of "5678" was actually assigned to `@product`, not `@product_id`.

**Workarea Setup (please complete the following information):**
Workarea v3.4

**Additional context**
This was originally reported by **John Varady** on discourse.

PR #146, opened 2019-10-03
`_id` Suffix Omitted From Customized Fields

tubbo:
When adding a customized field to a `Customizations` class that ends in
`_id`, Workarea was previously stripping this suffix from the computed
instance variable name that is converted into snake case from any kind
of input. This causes issues because the data doesn't appear to be
making it into customizations, but is really there under a different
instance variable name. To resolve the issue, Workarea now detects
whether a variable is already using snake case and leaves it
alone...only providing transformations for variable names that need it.

(#144)

### DIFF
--- a/core/app/models/workarea/catalog/customizations.rb
+++ b/core/app/models/workarea/catalog/customizations.rb
@@ -50,7 +50,7 @@ class Workarea::Catalog::Customizations
     @attributes = attributes.with_indifferent_access
 
     attributes.each do |name, value|
-      instance_variable_set("@#{name.to_s.titleize.systemize}", value)
+      instance_variable_set("@#{name.to_s.underscore.optionize('_')}", value)
     end
   end
 
--- a/core/test/models/workarea/catalog/customizations_test.rb
+++ b/core/test/models/workarea/catalog/customizations_test.rb
@@ -4,18 +4,22 @@ module Workarea
   module Catalog
     class CustomizationsTest < TestCase
       class FooCustomizations < Workarea::Catalog::Customizations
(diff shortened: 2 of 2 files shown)


core/test/models/workarea/catalog/customizations_test.rb
```diff
@@ -4,18 +4,22 @@ module Workarea
...
-          'titleGraphic' => 'gritty.jpg'
+          'titleGraphic' => 'gritty.jpg',
+          'category_id' => 'CAT',
+          'SCREAMING_SNAKE_ID' => '🐍'
         )
 
         assert_equal('Hello, World', customizations.a_test)
         assert_equal('gritty.jpg', customizations.title_graphic)
+        assert_equal('CAT', customizations.category_id)
+        assert_equal('🐍', customizations.screaming_snake_id)
```
meowsus: This snake doesn't look like it's screaming IMO

core/app/models/workarea/catalog/customizations.rb
```diff
@@ -50,7 +50,13 @@ def initialize(product_id, attributes)
     @attributes = attributes.with_indifferent_access
 
     attributes.each do |name, value|
-      instance_variable_set("@#{name.to_s.titleize.systemize}", value)
+      ivar = if name.to_s.include?('_')
```
bencrouse: Seems like we could just swap the `titleize` for `underscore` and achieve the same fix in a less brittle way? That `titleize` feels like it was the wrong idea in the first place.

tubbo: you're right, replacing it with underscore kept the tests passing.

PR #156, opened 2019-10-09
Rework FeaturedCategorizations to allow easier decoration

mttdffy:
WORKAREA-21

### DIFF
--- a/core/app/queries/workarea/featured_categorization.rb
+++ b/core/app/queries/workarea/featured_categorization.rb
@@ -3,8 +3,8 @@ module Workarea
     include Enumerable
     delegate :blank?, :present?, :size, :length, :to_a, to: :all
 
-    def initialize(product)
-      @product = product
+    def initialize(*product_ids)
+      @product_ids = product_ids
     end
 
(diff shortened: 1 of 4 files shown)


core/app/queries/workarea/featured_categorization.rb
```diff
@@ -39,9 +39,19 @@ def category_changesets_affecting_current_release
...
       FeaturedProducts
-        .changesets(@product.id)
+        .changesets(*product_ids)
         .where(releasable_type: Catalog::Category.name)
         .in(release_id: Release.current.preview.releases.map(&:id))
     end
+
+    private
+
+    def product_ids
```
bencrouse: If we're going to change this class to support multiple product IDs, and we don't really use the product model anyways, can we go all the way and update the initializer to accept multiple product IDs?

mttdffy: Its not meant for multiple products though, just that with some plugins, a single product can sometimes be represented with variations of it's ID. I'll bring this up in the standup

PR #158, opened 2019-10-10
Fix Payment cloning in Pricing::Request

mttdffy:
No changelog

### DIFF
--- a/core/app/models/workarea/pricing/request.rb
+++ b/core/app/models/workarea/pricing/request.rb
@@ -50,7 +50,7 @@ module Workarea
 
         @payment ||=
           Workarea::Payment.instantiate(@persisted_payment.as_document).tap do |payment|
-            payment.id = @persisted_payment.id # Ensure this isn't persisted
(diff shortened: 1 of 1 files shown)


core/app/models/workarea/pricing/request.rb
```diff
@@ -48,9 +48,10 @@ def shippings
       def payment
         return unless @persisted_payment.present?
 
-        @payment ||= @persisted_payment.clone.tap do |payment|
-          payment.id = @persisted_payment.id # Ensure this isn't persisted
-        end
+        @payment ||=
+          Workarea::Payment.instantiate(@persisted_payment.as_document).tap do |payment|
+            payment.id = @persisted_payment.id # Ensure this isn't persisted
```
bencrouse: I realized this actually still returns a savable document, this change will fix that:
```suggestion
            payment.new_record = true # Ensure this isn't persisted by raising duplicate key if saved
```

mttdffy: Wouldn't that allow it to save?

PR #161, opened 2019-10-11
Remove Puma Configuration from Install Generator

tubbo:
`workarea/configuration/puma` is no longer a file, so remove it from the
install generator and any mentions of it in the docs.


bencrouse:
@mcfearsome made this change to master, not intending to release as a patch. The Commerce Cloud plugin will work without removing this from base.

PR #162, opened 2019-10-11
Segment ux improvements

bencrouse:


bencrouse:
@meowsus I'd like to get this in ASAP so it's in @ebigler 's QA pass, can you review?

PR #166, opened 2019-10-14
Clean up Segement Rules UI

meowsus:
No changelog

WORKAREA-9

### DIFF
--- a/admin/app/views/workarea/admin/segment_rules/fields/_last_order.html.haml
+++ b/admin/app/views/workarea/admin/segment_rules/fields/_last_order.html.haml
@@ -1,3 +1,3 @@
-%span Ordered within
+%span Ordered in the last
 = number_field_tag 'rule[days]', rule.days.presence || Workarea.config.loyal_customers_last_order_days_ago, class: 'text-box text-box--small'
-%span days ago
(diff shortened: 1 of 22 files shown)


admin/app/views/workarea/admin/segment_rules/fields/_last_order.html.haml
```diff
@@ -1,3 +1,3 @@
-%span Ordered within
+%span= t('workarea.admin.segment_rules.fields.last_order.ordered_in')
```
bencrouse: I had left these untranslated for similar reason to the madlibs style in discounts (writing it in another language may not translate into the UI in the same way). So let's allow the variant to handle localizing these partials.

core/lib/workarea/visit.rb
```diff
@@ -57,6 +57,7 @@ def segments
     end
 
     def referrer
+      return {} if request.referrer.blank?
```
bencrouse: (to minimize object allocation)

```
def referrer
  return @referrer if defined?(@referrer)
  @referrer = Workarea.referrer_parser.parse(request.referrer) rescue {}
end
```

PR #169, opened 2019-10-15
Add "Implement the Credit Card Tender Type" doc

chriscressman:
* Update navigation and cross references
* Remove docs being replaced by this one
* Update "Test a Credit Card Transaction" doc

WORKAREA-22


chriscressman:
This procedure is the same for Workarea v3.4 and v3.5, so I decided to target v3.4 for the examples.

docs/source/articles/implement-the-credit-card-tender-type.html.md
```diff
@@ -0,0 +1,298 @@
...
+Workarea Braintree also decorates tests to configure the correct gateway and account for gateway API differences.
+Refer to the following sources:
+
+* [Workarea Core implementation](https://github.com/workarea-commerce/workarea/blob/v3.4.18/core/app/models/workarea/payment/store_credit_card.rb)
+* [Workarea Core test case](https://github.com/workarea-commerce/workarea/blob/v3.4.18/core/test/models/workarea/payment/store_credit_card_test.rb)
+* [Workarea Braintree implementation extensions](https://github.com/workarea-commerce/workarea-braintree/blob/v1.0.3/app/models/workarea/payment/store_credit_card.decorator)
+* [Workarea Braintree test case extensions](https://github.com/workarea-commerce/workarea-braintree/blob/v1.0.3/test/models/workarea/payment/store_credit_card_test.decorator)
+
+
+Operation Implementations
```
jyucis: You should make reference to the #cancel method that must be implemented. This method handles rolling back payment operations on error. This method also rolls back payment of fraud for some of our fraud plugins.

docs/source/articles/test-a-credit-card-transaction.html.md
```diff
@@ -1,16 +1,13 @@
...
-# Test a Credit Card Transaction
-
-[Active Merchant](http://activemerchant.org/) provides the following credit card numbers for use in development environments.
-
-| Card Number | Result |
-| --- | --- |
-| 1 | Successful tansaction |
-| 2 | Declined transaction |
-| 3 | Payment processing error |
+Test a Credit Card Transaction
```
jyucis: Maybe include something like: "Some credit card processors only accept specific credit card numbers in the test environment. Be sure to check your processors documentation."

docs/source/articles/implement-the-credit-card-tender-type.html.md
```diff
@@ -0,0 +1,298 @@
...
+  end
+
+  if transaction.response.success? && tender.token.blank?
+    tender.token =
+      transaction.response.params["braintree_transaction"]["credit_card_details"]["token"]
+    tender.save!
+  end
+end
+```
+
```
jyucis: One of the most common decorations in the operation classes is to decorate  #transaction_options. This method exists to pass customized data to the active merchant implementation.  I think its notable enough to mention here.

docs/source/articles/test-a-credit-card-transaction.html.md
chriscressman: I tried to make it more clear that this doc is referring to the behavior of the default bogus gateway only. This is a legacy doc that I thought was worth keeping, but cleaned it up a little.

docs/source/articles/implement-the-credit-card-tender-type.html.md
chriscressman: added it

docs/data/articles.yml
```diff
@@ -64,8 +64,7 @@ search:
   change-the-storefront-search-filters-ui.html.md: /articles/change-the-storefront-search-filters-ui.html.md
 cart-and-checkout:
   checkout.html.md: /articles/checkout.html.md
-  configure-a-payment-gateway.html.md: /articles/configure-a-payment-gateway.html.md
-  integrate-a-payment-gateway.html.md: /articles/integrate-a-payment-gateway.html.md
+  implement-the-credit-card-tender-type.html.md: /articles/implement-the-credit-card-tender-type.html.md
```
tubbo: could we shorten this to "Implement Credit Card Tenders"?

docs/source/articles/extension-overview.html.md
```diff
@@ -23,7 +23,7 @@ Workarea provides various _designed_ points of extension that employ conventions
 - [Content block types](add-a-content-block-type.html)
 - [Discount types](create-a-custom-discount.html)
 - Inventory policies
-- [Payment gateways](integrate-a-payment-gateway.html)
+- [Payment gateways](implement-the-credit-card-tender-type.html)
```
tubbo: name doesn't match title.

docs/source/articles/implement-the-credit-card-tender-type.html.md
```diff
@@ -0,0 +1,310 @@
...
+from Active Merchant, as the production gateway.
+The plugin also extends Active Merchant to provide its own bogus gateway, 
+[`ActiveMerchant::Billing::BogusBraintreeGateway`](https://github.com/workarea-commerce/workarea-braintree/blob/v1.0.3/lib/active_merchant/billing/bogus_braintree_gateway.rb).
+
+Workarea Braintree provides code to autoconfigure Workarea's credit card gateway.
+When Braintree credentials are present, the Braintree gateway is used, otherwise the bogus gateway is used
+([source](https://github.com/workarea-commerce/workarea-braintree/blob/v1.0.3/lib/workarea/braintree.rb#L8-L24)):
+
+```ruby
+def self.auto_configure_gateway
```
tubbo: I'm not a fan of including this anti-pattern in our documentation, even though it's currently how credit card gateways are implemented. The issue is that auto-configuring the gateway in an initializer doesn't work in multi-site if each site has different credentials (which is totally possible in an international setting). Either we need to draw a line in the sand and drop support for per-site credentials out-of-box, or we need to change the way that credit card gateways are instantiated so they can occur per-request, after the `Site.current` has been set. But we can't throw this out there and pretend like it works all the time, because SIs and clients constantly have trouble with this whole concept, especially when it gets down to the wire of launch date...

docs/source/articles/implement-the-credit-card-tender-type.html.md
```diff
@@ -0,0 +1,310 @@
...
+([source](https://github.com/workarea-commerce/workarea-braintree/blob/v1.0.3/app/models/workarea/payment/authorize/credit_card.decorator)):
+
+```ruby
+def complete!
+  transaction.response = handle_active_merchant_errors do
+    if tender.token.present?
+      gateway.authorize(
+        transaction.amount.cents,
+        tender.token,
+        {
```
tubbo: ~~these extra brackets are unnecessary, and many gateways implement these optional fields as keyword args anyway~~ never mind this is just a braintree example

docs/source/articles/test-a-credit-card-transaction.html.md
```diff
@@ -1,16 +1,15 @@
...
-| --- | --- |
-| 1 | Successful tansaction |
-| 2 | Declined transaction |
-| 3 | Payment processing error |
+By default, Workarea configures the credit card tender type to use a bogus gateway in development environments. This gateway, an instance of [`ActiveMerchant::Billing::BogusGateway`](https://www.rubydoc.info/gems/activemerchant/1.99.0/ActiveMerchant/Billing/BogusGateway), provides the following useful features for local development and testing:
 
+* Use credit card number ending in 1 for success, 2 for exception, and anything else for error
+* Use amount ending in 00 for success, 05 for failure, and anything else for exception
 
+If your application is using a different gateway in Development, consult the documentation for your gateway.
```
tubbo: We should expand this document to include VCR testing of credit card transactions. The bogus gateway and mocks are useful for developing other parts of the application, but not so much the actual mechanics of credit card processing. In fact, most (if not all?) of our credit card integration plugins set a real gateway and run VCR tests to ensure that making API calls will actually work in the real world.

At the very least, it should be noted that bogus gateways are _not_ actually testing your integration code, and you should use a real gateway for each credit card operation in tests, saving the responses to VCR..

docs/data/articles.yml
chriscressman: When all the docs in this set are finished the titles will be

* Payment Tender Types
* Implement the Credit Card Tender Type
* Implement a Primary Tender Type
* Implement an Advanced Payment Tender Type

with the first doc providing the context for the three that follow.

All four describe the "Tender Type" extension point provided by the platform.

docs/source/articles/extension-overview.html.md
chriscressman: Yeah, I should rename the extension point to "tender type". Since I don't have a generic "tender types" doc to link to yet, I'll continue linking to the credit card doc, but make it clear it is an example of a tender type.

docs/source/articles/implement-the-credit-card-tender-type.html.md
chriscressman: The Workarea team agreed this pattern should be updated in all the plugins that use it. We're going to do that work in a future sprint, after which I'll update this section with the updated examples.

chriscressman:
I'm going to finish all work on the tender types extension point in a single PR because the docs in the set have shared dependencies (the docs uncovered some needed code changes), and it will be easier for reviewers to understand where all the details are covered.

chriscressman:
Replaced by https://github.com/workarea-commerce/workarea/pull/217

PR #170, opened 2019-10-15
Fix missing aspect ratio magic attribute

bencrouse:
This magic attribute doesn't need to be calculated, it's the inverse of the aspect ratio we already have.

### DIFF
--- a/core/app/models/workarea/catalog/product_placeholder_image.rb
+++ b/core/app/models/workarea/catalog/product_placeholder_image.rb
@@ -45,6 +45,11 @@ module Workarea
         true
       end
 
+      def image_inverse_aspect_ratio
+        return if image_aspect_ratio.blank? || image_aspect_ratio.zero?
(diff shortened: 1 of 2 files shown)


core/app/models/workarea/catalog/product_placeholder_image.rb
```diff
@@ -44,6 +44,10 @@ def create_from_pipeline!
       def placeholder?
         true
       end
+      
+      def image_inverse_aspect_ratio
+        1 / image_aspect_ratio.to_f
```
mttdffy: should we ever have to worry about `#image_aspect_ratio` being `nil` and resulting in a divide by 0 error?

MMartyn: Good call, might as well return 0 if image_inverse_aspect_ratio is 0 or nil

PR #173, opened 2019-10-17
Implement Tribute.js for comment notifications

meowsus:
WORKAREA-6


admin/app/view_models/workarea/admin/commentable_view_model.rb
```diff
@@ -16,6 +16,10 @@ def comment_count
       def comments
         @comments ||= CommentViewModel.wrap(model.comments)
       end
+
+      def subscribed_users
+        @users ||= User.find(subscribed_user_ids)
```
bencrouse: This will raise if one of those IDs can't be found (e.g. the user was deleted). It'd be better to use `in(id: subscribed_user_ids)` which would return whoever it can find without raising.

admin/app/views/workarea/admin/comments/index.html.haml
```diff
@@ -44,21 +44,19 @@
...
+              = text_area_tag 'comment[body]', @comment.body, class: 'text-box text-box--multi-line', required: true, data: { tribute_area: users_path(exclude_current_user: true, format: :json) }
+              %span.property__note= t('workarea.admin.comments.index.mention_note')
 
-            .property
-              = label_tag 'subscribed_user_ids', t('workarea.admin.fields.subscribed_user_ids'), class: 'property__name'
-              = select_tag 'subscribed_user_ids', nil, multiple: true,  data: { remote_select: { source: users_path(exclude_current_user: true, format: :json) } }
-              %span.property__note= t('workarea.admin.comments.index.email_authentication')
+              - if @commentable.subscribed_users.any?
+                %strong= t('workarea.admin.comments.index.users_currently_subscribed')
+                = @commentable.subscribed_users.map(&:email).join(', ')
```
bencrouse: Not sure if it makes sense visually here, but I've used `to_sentence` instead of `join(', ')` in a couple spots recently and it was nice.

admin/app/views/workarea/admin/users/index.json.jbuilder
```diff
@@ -2,3 +2,11 @@ json.results @results do |user|
   json.label user.name
   json.value user.id
 end
+
+json.mentions @results do |user|
```
bencrouse: Any reason we can't combine the extra fields into the results above so it's one data set?

meowsus: A few reasons I did it this way:

- Create a dataset that serves a different module than Select2, since they require different data
- Make the module more direct, bypassing the need to munge the returned data into something usable for Tribute
- Be a more clear in intention of each data set

The `tributeAreas` module got a bit heftier than I originally anticipated because Tribute has no hooks for determining when a user changed their mind about a mention. A totally reasonable use-case would be if someone started typing a name, saw the name in the list, and accidentally clicked the wrong name. When they delete their mistake no events are triggered in tribute to announce that this has happened, so I had to write that into the `tributeAreas` module itself.

I felt that getting the data in the format that tribute needed was a low cost and high yield kind of move to save me from having to write another method in the module with high cognitive overhead.

PR #184, opened 2019-10-23
Remove HTTP caching headers for responses with segmented content

bencrouse:
If a page has segmented content, it can't be cached by any upstream HTTP
caches (including the browser) because the user's segments can change
request-by-request.

Our solution is to use the headers we've been using for cached responses
if the page has no segmented content. If it does have segmented content,
change those headers to force refetching every time.

This is being done in a last piece of middleware to ensure to
Rack::Cache the headers look the same. This allows us to still cache
complete responses in Rack::Cache for requests with segmented content.

This commit also refactors the middleware that sets this all up into a
single ApplicationMiddleware so it's easier to see everything going on
in one file.

WORKAREA-36


bencrouse:
Yeah, agreed. What's the upgrade entry? Like, the same 15 min rack cache still applies (segmented or not), and until you actually set segmented content, the headers will be the same as v3.4. So it won't actually require anyone to change things in the upgrade.

mttdffy:
I guess you're right. I was thinking of the upgrade guide more as a release notes thing I suppose.

PR #186, opened 2019-10-24
Don't default to S3 asset store

bencrouse:
This causes problems spinning up environments in other hosting setups
where S3 isn't available or desired.

To retain the old behavior (which you'll want if you're on the Workarea
Commerce Cloud) drop this into an initializer:
`Workarea.config.asset_store = (Rails.env.test? || Rails.env.development?) ?
:file_system : :s3`

WORKAREA-32


bencrouse:
@tubbo I'm not sure what you mean, the intent is to require explicitly setting `Workarea.config.asset_store = :s3` to get that functionality.

bencrouse:
Closing this, will open a new one for `master` based on conversations.

PR #189, opened 2019-10-25
Improve UX around Active By Segment tooltips

meowsus:
No changelog

WORKAREA-33


admin/app/assets/javascripts/workarea/admin/modules/active_by_segment_tooltips.js
```diff
@@ -58,36 +58,47 @@ WORKAREA.registerModule('activeBySegmentTooltips', (function () {
...
+                triggerClose: {
+                    mouseleave: true
                 }
             });
         },
 
+        setup = function (index, trigger) {
+            var $content = $(trigger.hash),
+                $select = $content.find('[data-select]'),
+                handleChange = _.partial(updateFields, $content, $(trigger));
```
tubbo: why is `$(trigger)` being passed into this function, but you don't seem to use it?

PR #190, opened 2019-10-26
[DOCS] Fix/improve various docs based on community feedback

chriscressman:


docs/source/articles/seeds.html.md
```diff
@@ -194,7 +194,7 @@ Refer to [`AdminSeeds#perform`](https://github.com/workarea-commerce/workarea/bl
 
 ## Re-Seeding
 
-The seeds included with Core use some random data, so seeding is not idempotent.&nbsp;<sup><a href="#notes" id="note-2-context">[1]</a></sup> However, re-seeding should produce data that is uniform—having the same general “shape”—each time.
+The seeds included with Core use some random data, so seeding is not idempotent. However, re-seeding should produce data that is uniform—having the same general “shape”—each time. (Of course, since seeds are extensible, you can write your own seeds to be idempotent if you so desire.)
```
bencrouse: How would writing idempotent seeds work? The task drops all Workarea collections each time...

chriscressman: I think it could be done with data files, but I also don't see much value in this aside. I moved it from a footnote, but I'll just remove it instead.

PR #193, opened 2019-10-28
Add device for segment rule

bencrouse:
This diff will make more sense after #184 is merged.


bencrouse:
I did an `ack` on projects, none of them were using `Robots` so this felt safe.

bencrouse:
@meowsus the other software doing similar don't bother, but I added a link the user agent wikipedia page which explains sniffing and spoofing.

PR #197, opened 2019-10-29
v3.4.20 Release Notes

tubbo:
Patch Release Notes for **v3.4.20**


docs/source/release-notes/workarea-3-4-20.html.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Workarea 3.4.17
```
bencrouse: There are 3 different versions in this file, can you update to the right one?

chriscressman:
Looks good (other than Ben's comment), but are there supposed to be notes for other patch versions?

PR #202, opened 2019-10-31
WIP

meowsus:


meowsus:
@bencrouse would you mind throwing this branch at your demo data to see if it fixes some of the issues you were seeing?

bencrouse:
@meowsus I attached a gif of how yesterday's demo data looks on this branch. note that there is one release from 9/16 that should be showing but isn't.

timeline_bug

PR #206, opened 2019-11-02
Clean up Timeline Report Chart UI

meowsus:
* Datasets contain points for every day
* Releases show regardless of other data in dataset
* Display dates in UTC, fixing date offset issue
* Re-align tooltip when hovering over Release datapoint

WORKAREA-31

### DIFF
--- a/admin/app/view_models/workarea/admin/reports/timeline_view_model.rb
+++ b/admin/app/view_models/workarea/admin/reports/timeline_view_model.rb
@@ -27,10 +27,23 @@ module Workarea
           }
         end
 
+        def releases
+          @releases ||= Release.published_between(
+            starts_at: starts_at,
+            ends_at: ends_at
+          ).to_a
+        end
+
         private
 
+        def date_range
+          starts_at.to_date..ends_at.to_date
+        end
+
         def grouped_data
-          results.group_by(&:starts_at)
+          date_range.each_with_object({}) do |date, group|
+            group[date] = results.select { |r| r.starts_at.to_date == date }
+          end
         end
 
         def data_for(type)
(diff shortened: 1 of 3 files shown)


meowsus:
The last thing I was trying to do was to customize the tooltips, but my previous approach wasn't working. To get the chart to display with the release indicators at the bottom of the page we need to force the `Y` axis value to `0`. Unfortunately there doesn't seem to be a way to send the original `Y` value data to the tooltip.

One idea I had was to list, next to the chart, dates that have releases with each release for that date in a sub-list underneath. There is probably a way that, when a user hovers over a release name, we highlight the date in the chart, producing the tooltip. Clicking on the release could take the user to the release show page.

Let's follow up with a side-convo or talk about it in the next UX meeting.

admin/app/view_models/workarea/admin/reports/timeline_view_model.rb
```diff
@@ -40,13 +46,14 @@ def data_for(type)
...
-            (values || []).map { |v| count_releases(v.starts_at.to_date) }
+          date_range.each_with_object({}) do |date, group|
+            data = releases.select { |r| r.published_at.to_date == date }
+            group[date] = [data.count]
           end
         end
 
         def transform(data)
-          data.map { |k, v| Hash[x: k, y: v.first] }.reverse
+          data.map { |k, v| Hash[x: k.to_time.utc, y: v.first] }.reverse
```
bencrouse: Does Chart.js just expect UTC?

meowsus: I can double check to see if this is still relevant. Chart.js uses Moment.js to do the conversion to local time on the fly, so that's why we were seeing the difference in dates. Since SalesByTime occurs at midnight, a `-700` offset put it one day behind in the chart.

PR #209, opened 2019-11-04
Fix Incorrect Test Setup

tubbo:
The `Pricing::Calculators::Calculator.test_adjust` method accepts two
arguments, and expects the first argument is going to be of type
`Order`, but in a `TaxCalculator` test only a shipping was being passed
in. Update this test to use the correct syntax so that other downstream
projects that expect data to be on an Order won't get confused.


tubbo:
if you do decide to merge this tomorrow morning, re-running the b2b build should fix its only failing test

PR #215, opened 2019-11-05
Remove now-unneeded version restriction on the BSON gem

bencrouse:
If we remove this restriction, we can use newer versions of the `mongo` gem, which contain cluster fixes.


bencrouse:
Everyone was cool with this when we talked yesterday, so I'm merging.

PR #207, opened 2019-11-04
Rename and update doc for testing CC transactions

chriscressman:
Rename "Test a Credit Card Transaction" to
"Manually Test Credit Card Transactions" and update the content.

The content was specific to a particular gateway and didn't make that
clear. These changes provide a generic solution in addition to the
specifics for the default gateway.

The title confused devs who reviewed this doc in a different PR,
thinking that it had to do with automated testing.

WORKAREA-13

PR #217, opened 2019-11-06
Add/improve payment tender types documentation

chriscressman:
Existing payment tender type documentation is limited to the credit card
tender type, covers only a portion of the implementation, and does not
explain the concepts or provide context to those new to the platform.

Replace and significantly expand coverage, providing 3 separate howtos:

* Customize the Credit Card Tender Type
* Implement a Primary Tender Type
* Implement an Advance Payment Tender Type

WORKAREA-13


chriscressman:
Replaces: https://github.com/workarea-commerce/workarea/pull/169

Related: https://github.com/workarea-commerce/workarea/pull/207

chriscressman:
So far, this PR replaces 2 old docs with 3 new howto docs and 1 new explain doc.

Each of the howtos has a section for operation implementations. The approach I took here is to provide boilerplate and then explain to the developer how they may need to change that boilerplate. But because there was so much to tell them about the boilerplate, I decided to use inline annotations to help them connect my comments about the boilerplate to the relevant lines of the boilerplate.

I think this code is different from other extension points in that we require the developer to write a large chunk of procedural code, but that code has to maintain a certain shape. Parts of it have to be changed and parts of it have to be maintained. For example, the dev has to assign a value to `transaction.response` and that value must be an object of a certain type.

In a future redesign of this code, we could design it so that the dev has to provide only the parts that are customizable, such as the object to be assigned, rather than making them do the assignment (which they cannot remove). We could also ensure the object they provide is of the right type.

There are many other examples of things that have to be present in their implementation. In the docs I refer to this as "fulfilling the operation implementation contract". It would be ideal to move the "contract" parts into code they don't change and provide the right hooks for the parts they are to change.

chriscressman:
I'm going to revise all pros and fill in missing stuff (TODOs) in future edits, so right now what I'm looking for is a technical review of the operation implementations sections, and any feedback you may have on the overall approach (described in my comment above).

docs/source/articles/implement-the-credit-card-tender-type.html.md
```diff
@@ -0,0 +1,1139 @@
...
+
+#### Boilerplate
+
+```ruby
+# your_engine/config/initializers/credit_card_gateway.rb [1][2]
+
+credentials = Rails.application.secrets.credit_card_gateway #[2][3]
+
+if credentials.present? #[4]
+  Workarea.config.gateways.credit_card = #[5]
```
chriscressman: Until the design of the credit card gateway extension point is changed in Workarea 3.6, I'm going to document the existing pattern of assigning a gateway instance to this config. We know this doesn't allow for different gateways for different sites in a multi site scenario, but we haven't encountered that being a problem for the credit card tender thus far.

I could write boilerplate that would work around this known issues in Workarea 3.5 and below, but there would be no concrete examples to show to the reader or for me to rely on to know that the solution works. I'd have to write my own solution and test it and then explain to the reader why the docs are different from what you see in actual plugins and application code.

But if anyone objects to my decision here, we can discuss it further.

chriscressman:
I added the "Gateway" section to the credit card tender howto doc, so I'm looking for technical review of that section in addition to the "Operation Implementations" sections across all 3 howto docs.

docs/source/articles/payment-tender-types.html.md
```diff
@@ -0,0 +1,54 @@
...
+
+_Explain_ concepts in this doc.
+So that devs working through a specific tender type implementation doc (credit card, primary, advance payment) can make sense of the steps.
+Provide context.
+
+
+Gateway
+----------------------------------------------------------------------
+
+Each tender type relies on a gateway.
```
bencrouse: Store credit doesn't, iirc

docs/source/articles/payment-tender-types.html.md
```diff
@@ -0,0 +1,54 @@
...
+It's often an instance of a class from the Active Merchant library, but...
+
+Some gateways are "bogus" gateways, which act like they're talking to a network service, but they use "canned" responses.
+Good for local development and manual testing because you can often control what response you get.
+For example, use a credit card ending in 1 to ensure a successful transaction or ending in 2 to get an exception.
+
+For most tender types, you want to initialize a different gateway for different use cases within the same application.
+For example, in production, you want to talk to the real network service, which requires initializing a gateway that knows how to do that, and initializing it with credentials.
+For local development, you may not care if you're talking to the real service, don't have credentials, and want the predictability and extra features of a bogus gateway.
+For automated testing, you might be fine using a bogus gateway, but for some or all tests you probably want to talk to the real service while developing the integration.
```
tubbo: This is more of a generic good practice than a "Workarea-specific" thing, but I think we could emphasize that there should be at least one test using VCR and hitting a real endpoint for each individual API call that you have to make, as the bogus gateway is typically not maintained as well as the "real" gateway. Furthermore, activemerchant integrations typically just use the `AM::BogusGateway` in development, which doesn't support extra methods or options needed by e.g. PayPal and others.

If you feel like this is better off in the "implementing a gateway" doc then that's cool too :-)

chriscressman:
I added the "Gateway"  and "Gateway Integration" sections to the advance payment tender types doc, so those are ready for technical review.

Update: I ported these sections to the primary tender types doc as well.

docs/source/articles/payment-tender-types.html.md
chriscressman: I'm defining _gateway_ as "the object responsible for processing payments of that type". So for store credit, the gateway is `Payment::Profile`. That's the only definition I can come up with that holds up across the major tender types we've implemented in base and in plugins.

chriscressman:
I added the "Credit Card Tokenization" section to the credit card type doc. That's ready for technical review.

chriscressman:
Merged v3.4-stable into this branch to get the visual improvements to the table of contents

chriscressman:
I applied all the changes we discussed in the docs meeting:

* Rewrite all examples to assume an Active Merchant gateway for credit cards and a non- Active Merchant gateway for primary and advance payment tender types.
* Replace the simple and full credit card boilerplate with a single boilerplate that includes both.
* Remove inline base implementations and examples. Link to them instead. Keep only boilerplate as inline code so that it's easier for copy-pasters to figure out what they want to copy/paste.
* Replace PayPal examples with Afterpay examples.
* Generally revise the credit card code annotations.

chriscressman:
Merged v3.4-stable into this branch to get the visual improvements to links containing inline code

chriscressman:
I added to the primary tender types doc and the advance payment tender types doc the sections for the tender type definition and the payment integration

chriscressman:
I revised the payment integration section for primary tender types because I discovered some additional steps that are necessary in all cases.

chriscressman:
I added the storefront and admin integration sections, so the 3 howto docs are ready for full technical review. Feel free to comment here, or we can discuss at the next docs meeting. I'm going to begin going over all 3 howtos to write the final pros.

chriscressman:
I removed the explain doc for now. I'll have to move that into another chunk of work.

chriscressman:
Credit card tender type howto doc is up and ready for final review.

chriscressman:
Changes/additions I'd like to make in an upcoming sprint (separate PR):

* Revise the credit card tender howto doc to use the same outline format as the other two docs
  * Present the task as a "procedure of procedures"
  * This is a prerequisite for later decomposing this set into many smaller docs
* Add an "explain" doc that puts these procedures into the wider context of checkout and payment and explains the domain concepts (e.g. primary tender, advance payment tender, transaction, operation)
  * Ideally, move existing explanations within the howto docs out of those docs and into the explain doc
* Add screenshots to support the Storefront Integration sections (either directly in those sections or in the explain doc)

Changes/additions I'd like to make in a future iteration (e.g. after Workarea 3.6):

* Update the docs to reflect changes in Workarea 3.6 (after those changes are merged)
* Change the Operation Implementations subsections to NOT "share" annotations
  * These are the only sections written like this and they can't stand on their own when written like this
  * This is a prerequisite for decomposing the docs (below)
* Decompose the entire document set, resulting in a larger set of shorter docs
  * As part of this effort, try to generalize some of the docs so they can be shared (DRY up the docs)
* Cover testing for each of the documented procedures
  * Testing is barely covered right now, but it will add a lot of length, and the docs are already unbearably long
* Update the Storefront Integration and Admin Integration sections to use the "boilerplate + example" format the other sections use
  * This provides the best user experience for the reader
  * And it will make the docs internally consistent
* Improve the Tender-Specific Integration sections by replacing lists of source files with descriptions of the integrations (with relevant links to sources) and providing examples for the Admin

chriscressman:
I squashed and rebased. I changed the base branch, since docs now publish from master.

This is ready for final review and merge.

bencrouse:
Gonna wait on a few more eyeballs before merging.

PR #219, opened 2019-11-06
Fix and update traffic referrer segment rule

bencrouse:
This makes it more consistent with other rules like geolocation and
browser info. It also allows regex matching on any referrer URL, whether
it's known or not.

No changelog


admin/app/views/workarea/admin/segment_rules/summaries/_traffic_referrer.html.haml
```diff
@@ -1,10 +1,2 @@
...
-    %span #{t('workarea.admin.fields.medium')}:
-    %strong= rule.medium.titleize
-    ,
-
-- if rule.source.present?
-  %span.text.text--large
-    %span #{t('workarea.admin.fields.source')}:
-    %strong= rule.source
+%span.text.text--large
+  %strong= [rule.medium.try(:titleize), rule.source, rule.url].flatten.reject(&:blank?).to_sentence
```
tubbo: why not `.compact` instead of `.reject(&:blank?)`?

bencrouse: `reject(&:blank?)` will get rid of blank strings

PR #221, opened 2019-11-08
Add index to SearchByDay model

balexandr:
I noticed in one of our projects, SearchCustomizations and SearchSettings were slugging along. I discovered that SearchByDay was the root cause of the slowness/timeouts. SearchByWeek had the index so I think it may have just been an oversight. Let me know if it was intentionally left out.


bencrouse:
Awesome, thanks @balexandr!

PR #232, opened 2019-11-14
Fix segment content resetting in tests

bencrouse:
Also refactor setup/teardown so we can share that in one spot.

No changelog

### DIFF
--- a/testing/lib/workarea/test_case.rb
+++ b/testing/lib/workarea/test_case.rb
@@ -208,8 +208,34 @@ module Workarea
       end
     end
 
+    module Setup
+      extend ActiveSupport::Concern
+
+      included do
+        setup do
+          Mongoid.truncate!
+          Workarea.redis.flushdb
+          WebMock.disable_net_connect!(allow_localhost: true)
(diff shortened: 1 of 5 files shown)


testing/lib/workarea/test_case.rb
```diff
@@ -208,8 +208,34 @@ def reset_encryption_key
       end
     end
 
+    module Setup
```
bencrouse: If anyone can think of a word to represent both setup and teardown, I'm into combining these into a single module.

tubbo: `TestCase::Environment` doesn't necessarily represent the two verbs you mentioned, but IMO it's what this module does: sets up and tears down the _environment_ used to run tests. You could also call it `TestCase::Harness` in relation to the alternative definition here: https://en.wikipedia.org/wiki/Test_harness

tubbo: also thank you for doing this, i remember going down a rabbit hole trying to add mailer tests once and i realized i had to do a lot of copy pasta

PR #234, opened 2019-11-15
Allow redis to be configured with TLS

MMartyn:


MMartyn:
I opted to let the redis client parse the pieces so that I could optionally include a password in the host env var, but I could also switch this to having another env var for redis password if that would be preferred.

https://github.com/redis/redis-rb/blob/master/lib/redis/client.rb#L403

core/lib/workarea/configuration/redis.rb
```diff
@@ -67,8 +71,14 @@ def db
         @config[:db]
       end
 
+      def to_h
```
bencrouse: I'm curious, why switch this to be only a URL? How would I get a hash-version of the config now?

bencrouse: Ah, I missed your comment above. I think for compatibility it'd be best to keep the hash version around, can you expand the hash to include all the things?

bencrouse:
@MMartyn hey if you want to get this into v3.5, you'll need to update this week.

core/lib/workarea/configuration/redis.rb
MMartyn: Updated

core/lib/workarea/configuration/redis.rb
```diff
@@ -39,11 +39,15 @@ def find_config(name)
...
+          scheme = ENV["WORKAREA_#{env_slug}_SCHEME"].presence || DEFAULT[:scheme]
 
           {
+            scheme: scheme,
             host: ENV["WORKAREA_#{env_slug}_HOST"].presence || DEFAULT[:host],
             port: ENV["WORKAREA_#{env_slug}_PORT"].presence || DEFAULT[:port],
-            db: ENV["WORKAREA_#{env_slug}_DB"].presence || DEFAULT[:db]
+            db: ENV["WORKAREA_#{env_slug}_DB"].presence || DEFAULT[:db],
+            password: ENV["WORKAREA_#{env_slug}_PASSWORD"].presence,
+            ssl: scheme == 'rediss' ? true : false
```
bencrouse:
```suggestion
            ssl: scheme == 'redis' ? true : false
```

bencrouse: @MMartyn I can't commit this suggestion since its your personal fork, can you commit that?

MMartyn: I can see how it might seem like a typo, but rediss was intentional.

PR #236, opened 2019-11-18
Add special tags information tooltip to content asset tags field

mttdffy:
WORKAREA-99

### DIFF
--- a/admin/config/locales/en.yml
+++ b/admin/config/locales/en.yml
@@ -746,6 +746,20 @@ en:
           save_asset: Save Asset
           tags_label: Tags
           tags_note: 'Comma separated: just, like, this'
+          tags_tooltip_html: |
+            <p>Use the following tags to set an asset as a favicon or Open Graph image for your site:</p>
+
+            <ul>
+              <li><code>favicon</code> - The default favicon</li>
+              <li><code>favicon-WxH</code>, e.g. "favicon-32x32" - The favicon for the given size</li>
+              <li><code>og-default</code> - The default Open Graph image</li>
+            </ul>
+
+            <p>
+              The default favicon is used when a device does not request a specific size or when the requested size is not available.
+              <br/>
+              The default Open Graph image is used when sharing a content page that does not specify an Open Graph image.
+            </p>
           type_label: Type
         flash_messages:
           created: Asset has been created.
--- a/admin/app/views/workarea/admin/content_assets/edit.html.haml
+++ b/admin/app/views/workarea/admin/content_assets/edit.html.haml
@@ -24,9 +24,14 @@
           = text_field_tag 'asset[name]', @asset.name, class: 'text-box'
 
         .property
-          = label_tag 'asset_tag_list', t('workarea.admin.fields.tags'), class: 'property__name'
+          = label_tag 'asset_tag_list', class: 'property__name' do
+            %span= t('workarea.admin.fields.tags')
+            = link_to '#tags-info', data: { tooltip: '' } do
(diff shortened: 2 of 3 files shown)


admin/config/locales/en.yml
```diff
@@ -746,6 +746,21 @@ en:
...
+              <strong>Favicons:</strong> If you want to use an image asset as a favicon for the site,
+              you can add the <code>favicon</code> tag. Additionally, you can specify use for a
+              specific size favicon by tagging it with <code>favicon-HxW</code>. For example, a tag of
+              <code>favicon-32x32</code> will indicate the asset should be used for the favicon when
+              it's size is 32x32.
+            </p>
+            <p>
+              <strong>Open Graph:</strong> You can specify what image asset to use by default for the
+              open graph image of your content pages by tagging it with the <code>og-default</code> tag.
+              When shared on Facebook, that asset will display if there is no open graph image specified
```
bencrouse:
```suggestion
              When shared on social media, that asset will display if there is no open graph image specified
```

admin/config/locales/en.yml
```diff
@@ -746,6 +746,21 @@ en:
...
           tags_label: Tags
           tags_note: 'Comma separated: just, like, this'
+          tags_tooltip_html: |
+            <p>You can specify tags on an asset to mark them for special use cases.</p>
+            <p>
+              <strong>Favicons:</strong> If you want to use an image asset as a favicon for the site,
+              you can add the <code>favicon</code> tag. Additionally, you can specify use for a
+              specific size favicon by tagging it with <code>favicon-HxW</code>. For example, a tag of
+              <code>favicon-32x32</code> will indicate the asset should be used for the favicon when
+              it's size is 32x32.
```
bencrouse: Similar to explaining sharing below, should we briefly explain why you might use the different sizes?

admin/config/locales/en.yml
```diff
@@ -746,6 +746,21 @@ en:
           save_asset: Save Asset
           tags_label: Tags
           tags_note: 'Comma separated: just, like, this'
+          tags_tooltip_html: |
+            <p>You can specify tags on an asset to mark them for special use cases.</p>
+            <p>
+              <strong>Favicons:</strong> If you want to use an image asset as a favicon for the site,
```
bencrouse:
```suggestion
              <strong>Favicons:</strong> If you want to use an image asset as the default favicon for the site,
```

admin/config/locales/en.yml
mttdffy: is this true? I thought Facebook was the only use for open graph tags.

admin/config/locales/en.yml
```diff
@@ -746,6 +746,21 @@ en:
           save_asset: Save Asset
           tags_label: Tags
           tags_note: 'Comma separated: just, like, this'
+          tags_tooltip_html: |
+            <p>You can specify tags on an asset to mark them for special use cases.</p>
```
chriscressman: I'm not sure what markup is best of this context, but regarding content, how about something like this?:

Use the following tags to set an asset as a favicon or Open Graph image for your site:

"favicon" - The default favicon
"favicon-<var>W</var>x<var>H</var>", e.g. "favicon-32x32" - The favicon for the given size
"og-default" - The default Open Graph image

The default favicon is used when a device does not request a specific size or when the requested size is not available.
The default Open Graph image is used when sharing a content page that does not specify an Open Graph image.

admin/config/locales/en.yml
bencrouse: I know Slack also uses it, maybe others.

admin/config/locales/en.yml
bencrouse: sounds good to me.

PR #237, opened 2019-11-18
Improve presentation of code within links

chriscressman:
The presentation of inline code (e.g. class names) within hyperlinks is
a bit jarring. Only recent docs are linking to code, so it is just now
appearing as an issue.

Adjust the styles for code within links to be more traditional.

WORKAREA-108

No changelog


chriscressman:
Before:

before

After:

after

PR #238, opened 2019-11-18
Add Custom Events to Timeline Report

meowsus:
WORKAREA-86

### DIFF
--- a/admin/app/controllers/workarea/admin/custom_events_controller.rb
+++ b/admin/app/controllers/workarea/admin/custom_events_controller.rb
@@ -1,0 +1,40 @@
+module Workarea
+  module Admin
+    class CustomEventsController < Admin::ApplicationController
+      before_action :find_event
+
+      def create
+        flash[:success] = t('workarea.admin.reports.timeline.flash_messages.created')
+        redirect_back fallback_location: root_path
+      end
+
+      def update
+        if @custom_event.update_attributes(params[:custom_event])
+          flash[:success] =
+            t('workarea.admin.reports.timeline.flash_messages.success')
+        else
+          flash[:error] =
+            t('workarea.admin.reports.timeline.flash_messages.error')
+        end
+
+        redirect_back fallback_location: root_path
+      end
+
+      def destroy
+        @custom_event.destroy
+        flash[:success] = t('workarea.admin.reports.timeline.flash_messages.success')
+        redirect_back fallback_location: root_path
+      end
+
+      private
+
+      def find_event
+        @custom_event = if params[:id].present?
+          Workarea::Reports::CustomEvent.find_by(id: params[:id])
+        else
+          Workarea::Reports::CustomEvent.create!(params[:custom_event])
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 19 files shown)


meowsus:
I'm working on the tests, but want to make sure any changes in the PR won't bone my testing up.

Screenshot from 2019-11-18 11-49-30

admin/app/controllers/workarea/admin/custom_events_controller.rb
```diff
@@ -0,0 +1,38 @@
+module Workarea
+  module Admin
+    class CustomEventsController < Admin::ApplicationController
+      before_action :find_event, except: :create
```
bencrouse: So the `@custom_event` is only used in `update` so this doesn't really make sense to run before every action in its current state. You could either separate this out and not use the `before_action` or, rewrite `find_event` to do something that works for all actions, like
```
def find_event
  @event = if params[:id].present?
    Reports::CustomEvent.find(params[:id])
  else
    Reports::CustomEvent.new(params[:event])
  end
end
```

bencrouse: edit - maybe it was needed was for views? In which case, you could just do the above and wrap the result in the view model.

admin/app/view_models/workarea/admin/reports/timeline_view_model.rb
```diff
@@ -4,37 +4,74 @@ module Reports
       class TimelineViewModel < ApplicationViewModel
         include GroupByTime
 
+        class Event
```
bencrouse: Hey I probably didn't communicate too well, so I started off some cleanup on this to hopefully clarify. I haven't tested this code! Happy to talk through the notes.

```ruby
class Event
  def self.build(releases, custom_events)
    # this will automatically use a new Array as the default value for each key
    result =  Hash.new { |h, k| h[k] = [] } 

    # Handling these separately will allow handling the differences pretty simply
    releases.each do |release|
      result[release.published_at.to_date] << new(ReleaseViewModel.wrap(release), type: 'release')
    end

    custom_events.each do |event|
      result[event.occured_at.to_date] << new(event, type: 'custom_event')
    end

    # Sorting and convert back to hash so you sort by the key
    result.sort.to_h
  end

  delegate_missing_to :@model
  attr_reader :model, :type

  def initialize(model, type:)
    @model = model
    @type = type
  end
end
```

core/app/models/workarea/release.rb
```diff
@@ -12,6 +12,8 @@ class Release
     field :undone_at, type: Time # TODO deprecated, remove in v3.6
     field :undo_job_id, type: String # TODO deprecated, remove in v3.6
 
+    alias_method :occurred_at, :published_at
```
bencrouse: I think this might not be needed after cleanup I started above?

meowsus:
> Can we get some integration test coverage over the CRUD on the custom events? You thinking a system test over the report in general is possible?

@bencrouse The tests are incoming. I think I'll be able to get a good amount of system tests in place, especially for the new Event functionality. I just wanted to make sure nothing shook out from the preliminary code review that would affect my testing negatively.

PR #239, opened 2019-11-19
Update Tests Referencing 2020

### LINKED ISSUE #222: Tests to fail in 2020
Some of these tests will fail in 2020 (credit card expirations)

https://github.com/workarea-commerce/workarea/search?q=2020&unscoped_q=2020

tubbo:
The credit card expiration year `2020` was hard-coded into many Workarea
integration tests, and would fail when January 2020 passes. Update these
tests to always set the credit card expiration year to 3 years in
advance of when the test runs so this won't happen again in the future.

WORKAREA-104

Fixes #222

### DIFF
--- a/storefront/test/integration/workarea/storefront/checkout_side_effects_integration_test.rb
+++ b/storefront/test/integration/workarea/storefront/checkout_side_effects_integration_test.rb
@@ -74,7 +74,7 @@ module Workarea
         assert_equal('Crouse', credit_card.last_name)
         assert_equal('XXXX-XXXX-XXXX-1', credit_card.display_number)
         assert_equal(1, credit_card.month)
-        assert_equal(2020, credit_card.year)
+        assert_equal(next_year, credit_card.year)
         refute(credit_card.token.blank?)
 
         complete_checkout('bcrouse@workarea.com', 'W3bl1nc!')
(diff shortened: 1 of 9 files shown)


storefront/app/views/workarea/storefront/style_guides/components/_data_card.html.haml
```diff
@@ -1,3 +1,4 @@
+- year = Time.current.year + 3
```
mttdffy: dont _really_ need to change this.

testing/lib/workarea/test_case.rb
```diff
@@ -169,13 +169,26 @@ def reset_s3
       end
     end
 
+    module MorrisDayAndTheTime
```
bencrouse: What is this? Can we stick with descriptive naming?

testing/lib/workarea/test_case.rb
```diff
@@ -169,13 +169,26 @@ def reset_s3
...
+      extend ActiveSupport::Concern
+
+      included do
+        teardown :travel_back
+      end
+
+      def in_three_years
```
bencrouse: Why not `next_year` => `1.year.from_now.year`? Is there something special about 3 years?

testing/lib/workarea/test_case.rb
meowsus: I lol'd

tubbo: image

I honestly didn't know what to call it. Ran out of names yesterday.

tubbo:
> I think the module should be mixed into all test types. Also, let's fix this in `v3.4-stable`

I made it a factory method instead so that will work. Also changed the branch, that was my bad.

PR #240, opened 2019-11-20
Ensure UTF-8 Encoding of Data File Import Samples

tubbo:
This ensures data file import samples are always treated as UTF-8.
While Ruby itself does do this pretty well, and most browsers are good
at guessing the file type/encoding based on the contents of the file,
there might be some outliers that rely on metadata that's a bit more
strictly adhered to. This change ensures that sample CSV/JSON files are
delivered to the user as an attachment, and using the correct MIME type,
so that they register as such when downloaded by the browser.
Previously, all imports were showing as "TXT file" types, when they were
really "CSV file" or "JSON file", and at least in Firefox, they were not
downloading when you clicked the sample link. Instead, a new tab would
open (since Firefox thinks it's a text file), and you have to refresh
the page to actually get the browser to download the file.

### DIFF
--- a/admin/app/controllers/workarea/admin/data_file_imports_controller.rb
+++ b/admin/app/controllers/workarea/admin/data_file_imports_controller.rb
@@ -8,7 +8,7 @@ module Workarea
 
       def sample
         filename = "#{@import.name.downcase.underscore}.#{@import.file_type}"
-        send_data @import.sample_file_content, filename: filename, type: @import.file_type
+        send_data @import.sample_file_content, filename: filename, type: @import.mime_type, disposition: :attachment
       end
 
       def create
--- a/core/app/models/workarea/data_file/operation.rb
+++ b/core/app/models/workarea/data_file/operation.rb
@@ -37,6 +37,10 @@ module Workarea
           Workarea.config.data_file_formats.first
       end
 
+      def mime_type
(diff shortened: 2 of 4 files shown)


admin/app/controllers/workarea/admin/data_file_imports_controller.rb
```diff
@@ -8,7 +8,7 @@ def new
 
       def sample
         filename = "#{@import.name.downcase.underscore}.#{@import.file_type}"
-        send_data @import.sample_file_content, filename: filename, type: @import.file_type
+        send_data @import.sample_file_content, filename: filename, type: @import.mime_type, disposition: :attachment
```
bencrouse: Can we get an integration test checking the correct headers?

tubbo:
made sure that setting `bom|utf-8` as the default encoding works for UTF-8 files with a BOM and without a BOM. seems OK!

core/app/models/workarea/data_file/operation.rb
```diff
@@ -37,6 +37,10 @@ def file_type
           Workarea.config.data_file_formats.first
       end
 
+      def mime_type
+        "#{MIME::Types.type_for(file_type).first.to_s}; charset=utf-8"
```
bencrouse: This is a nice refactor compared with before

PR #241, opened 2019-11-20
Add admin alert when a newer workarea version is available

mttdffy:
WORKAREA-107

### DIFF
--- a/admin/config/locales/en.yml
+++ b/admin/config/locales/en.yml
@@ -2375,7 +2375,7 @@ en:
           product: product
           publishes_on_and_has_no_changes: "%{release_name} publishes on %{release_publish_date} and has no changes"
           variant: variant
-          workarea_version_outdated: A newer version of Workarea is availble! Contact your system integration team about upgrading.
+          workarea_version_outdated: A newer version of Workarea is available! Contact your implementer about upgrading.
         avatar_title: Avatar for %{name}
         browse_storefront_as_guest: Browse the Storefront as a Guest
         close_primary_menu: Close Primary Menu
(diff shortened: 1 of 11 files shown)


admin/config/locales/en.yml
```diff
@@ -2375,6 +2375,7 @@ en:
           product: product
           publishes_on_and_has_no_changes: "%{release_name} publishes on %{release_publish_date} and has no changes"
           variant: variant
+          workarea_version_outdated: A newer version of Workarea is availble! Contact your system integration team about upgrading.
```
bencrouse:
```suggestion
          workarea_version_outdated: A newer version of Workarea is available! Contact your implementer about upgrading.
```

core/lib/workarea/latest_version.rb
```diff
@@ -0,0 +1,25 @@
+module Workarea
+  module LatestVersion
+    def self.get
+      Rails.cache.fetch('workarea/version/latest', expires_in: 3.days) do
+        ENV['HTTP_PROXY'] || ENV['HTTPS_PROXY']
```
bencrouse: Is this just leftover cruft? `Net::HTTP` will pay attention to these, so I believe it isn't necessary to add them to the request.

mttdffy: I based this on https://github.com/workarea-commerce/workarea/blob/master/core/lib/workarea/ping_home_base.rb#L9 and assumed there was some wizardry behind invoking the variables beforehand. If its cruft in both places I can remove

bencrouse: It's gotta be cruft, go for it.

PR #243, opened 2019-11-20
Import UTF-8 CSVs With BOM Characters

tubbo:
UTF-8 doesn't need a BOM in order to start or end a file, but these
characters can end up in CSVs generated by older software that doesn't
have great support for Unicode. As a result, if a BOM is in the CSV near
`_id` it will cause improper importing of the data held within. To
address this, Workarea now specifies the `bom|` prefix in the `:encoding`
param for `CSV.foreach` by default. This can still be overridden if you
have an ASCII file, and since BOM stripping doesn't really apply,
developers can override the entire encoding string in configuration
if necessary. But this is a sane default for those who use UTF-8 and
happen to be exporting out of older spreadsheet software.

### DIFF
--- a/core/test/models/workarea/data_file/csv_test.rb
+++ b/core/test/models/workarea/data_file/csv_test.rb
@@ -403,6 +403,21 @@ module Workarea
           assert_equal("Bowl Light\u0099", product.name)
         end
       end
+
+      def test_bom_characters_in_unicode
+        csv = %(\xEF\xBB\xBF_id,name,slug,details_keywords\n653911,Bowl Light,Bowl_Light,"testing")
+        import = create_import(
+          model_type: Catalog::Product.name,
+          file: create_tempfile(csv, extension: 'csv'),
+          file_type: 'csv'
+        )
+
+        Csv.new(import).import!
+
+        product = Catalog::Product.find_by(slug: 'bowl_light')
+
+        assert_equal('653911', product.id)
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


core/test/models/workarea/data_file/csv_test.rb
```diff
@@ -403,6 +403,21 @@ def test_encoding_conversion
           assert_equal("Bowl Light\u0099", product.name)
         end
       end
+
+      def test_bom_characters_in_unicode
+        csv = %(\xEF\xBB\xBF_id,name,slug,details_keywords\n653911,Bowl Light,Bowl_Light,"testing")
```
meowsus: [B'Owl](https://www.youtube.com/watch?v=KweD1Bxjt9w)

mttdffy: [bom chicka wow ow](https://www.youtube.com/watch?v=vgxxAwue7Fs)

core/app/models/workarea/data_file/csv.rb
```diff
@@ -5,6 +5,7 @@ def import!
         index = 2 # start at 1 and skip headers
         failed_new_record_ids = []
         options = Workarea.config.csv_import_options.merge(headers: true)
+        options[:encoding] ||= 'bom|utf-8'
```
bencrouse: Why wouldn't this be in the default value of `Workarea.config.csv_import_options` instead of setting here?

PR #245, opened 2019-11-20
Base discount auto-deactivation on updated_at, expose auto_deactivate field

mttdffy:
WORKAREA-114

### DIFF
--- a/admin/app/helpers/workarea/admin/application_helper.rb
+++ b/admin/app/helpers/workarea/admin/application_helper.rb
@@ -156,6 +156,16 @@ module Workarea
         new_query_string_params = request.query_parameters.merge(page: page)
         "#{request.path}?#{new_query_string_params.to_query}"
       end
+
+      def duration_in_words(duration)
+        parts = duration.parts
+        return t('workarea.duration.seconds', count: 0) if parts.empty?
+
+        parts
+          .sort_by { |unit,  _ | ActiveSupport::Duration::PARTS.index(unit) }
+          .map     { |unit, val| t("workarea.duration.#{unit}", count: val) }
+          .to_sentence
+      end
     end
   end
 end
--- a/admin/app/views/workarea/admin/pricing_discounts/edit.html.haml
+++ b/admin/app/views/workarea/admin/pricing_discounts/edit.html.haml
@@ -30,11 +30,22 @@
           = text_field_tag  'discount[tag_list]', @discount.tag_list, class: 'text-box'
           %span.property__note= t('workarea.admin.form.csv_field_note')
 
-        .property
-          %span.property__name
-            = t('workarea.admin.fields.active')
-            = inline_svg('workarea/admin/i18n.svg', class: 'svg-icon svg-icon--small') if locale_options.many? && Workarea.config.localized_active_fields
-          = toggle_button_for 'discount[active]', @discount.active?
+        .grid
+          .grid__cell.grid__cell--50.grid__cell--25-at-medium
+            .property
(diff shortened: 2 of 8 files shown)


admin/app/views/workarea/admin/pricing_discounts/edit.html.haml
```diff
@@ -30,11 +30,22 @@
...
+              = toggle_button_for 'discount[active]', @discount.active?
+          .grid__cell.grid__cell--50.grid__cell--25-at-medium
+            .property
+              %span.property__name
+                = t('workarea.admin.fields.auto_deactivate')
+                = link_to '#auto-deactivate-info', data: { tooltip: '' } do
+                  = inline_svg('workarea/admin/icons/help.svg', class: 'svg-icon svg-icon--small svg-icon--blue')
+              = toggle_button_for 'discount[auto_deactivate]', @discount.auto_deactivate?
+
+            #auto-deactivate-info.tooltip-content= t('workarea.admin.pricing_discounts.edit.auto_deactivate_info', duration: Workarea.config.discount_staleness_ttl.inspect)
```
bencrouse: Is the `inspect` localized here? Seems better to use `distance_of_time_in_words`.

mttdffy: `distance_of_time_in_words` is for dates and times. This is a duration.

bencrouse: It works for both:
```
irb(main):001:0> helper.distance_of_time_in_words(1.month)
=> "about 1 month"
```

Not a big deal if `inspect` localizes, but I'd guess that it doesn't, I'm not sure.

mttdffy: the "about" isn't really desirable". 

> Discounts that auto-deactivate will become inactive when not updated or redeem for about 30 days

I'll see what other options there are

bencrouse: Oh I see. https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html

Check out the `scope` option. We could do a scope called `short` or `precise` or something that drops the "about", and then don't need a custom method.

```
irb(main):006:0> helper.distance_of_time_in_words(1.month, 0, scope: 'precise')
=> "translation missing: en.precise.about_x_months"
```

mttdffy: did you see that I added a helper to handle making it localized? scope doesn't work.

admin/app/helpers/workarea/admin/application_helper.rb
```diff
@@ -156,6 +156,16 @@ def pagination_path_for(page: 1)
         new_query_string_params = request.query_parameters.merge(page: page)
         "#{request.path}?#{new_query_string_params.to_query}"
       end
+
+      def duration_in_words(duration)
```
bencrouse: This would make a good PR to Rails

admin/app/views/workarea/admin/pricing_discounts/edit.html.haml
bencrouse: I was thinking you'd define the translation as:
```yaml
en:
  distance_of_time_in_words:
    short:
      about_x_months:
        one: '1 hour'
        other: '%{count} hours'
```
and call as `distance_of_time_in_words(1.month, 0, scope: 'distance_of_time_in_words.short')`

PR #250, opened 2019-11-21
Clean up Timeline Report Chart

meowsus:
QA uncovered some weirdness, namely:

* Stop Orders from being active by default
* Stop bubbling of the events triggered from within the tooltip
* Order events by time when many are displayed within the same day
* Display Events in graph above other data sets
* Fix tooltip functionality browsers other than Chrome

No changelog

WORKAREA-86

### DIFF
--- a/admin/app/view_models/workarea/admin/reports/timeline_view_model.rb
+++ b/admin/app/view_models/workarea/admin/reports/timeline_view_model.rb
@@ -11,14 +11,16 @@ module Workarea
             releases.each do |release|
               result[release.published_at.to_date] << new(
                 Workarea::Admin::ReleaseViewModel.wrap(release),
-                type: 'release'
+                type: 'release',
+                occurred_at: release.published_at
               )
             end
 
             custom_events.each do |event|
               result[event.occurred_at.to_date] << new(
                 event,
-                type: 'custom_event'
+                type: 'custom_event',
+                occurred_at: event.occurred_at
               )
             end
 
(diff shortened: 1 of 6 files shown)


admin/app/view_models/workarea/admin/reports/timeline_view_model.rb
```diff
@@ -49,18 +54,26 @@ def graph_data
...
-              releases: transform(graph_data_for('releases')),
-              custom_events: transform(graph_data_for('custom_events'))
+              customers: transform(graph_data_for('customers'))
             }
           }
         end
 
         def events
-          @events ||= Event.build(releases, custom_events)
+          @events ||= begin
```
bencrouse: Would you add a unit test to cover this logic?

admin/app/view_models/workarea/admin/reports/timeline_view_model.rb
```diff
@@ -32,6 +32,11 @@ def initialize(model, type:)
             @model = model
             @type = type
           end
+
+          def occurred_at
+            return @model.published_at if @type == 'release'
```
bencrouse: One way around this awkwardness would be to pass in the `occured_at` when creating `Event.new`

mttdffy: I wouldn't mind just aliasing `published_at` to `occured_at`. 🦆

PR #255, opened 2019-11-25
Add caching to order item details results

bencrouse:
This was a point of bottleneck during recent Reformation load-testing.

WORKAREA-102


bencrouse:
Yeah def. Not sure how else to help.

PR #263, opened 2019-12-02
Add activate_with as a field on block drafts for compatibility

mttdffy:


core/app/models/workarea/content/block_draft.rb
```diff
@@ -8,6 +8,10 @@ class BlockDraft
         field name, field_instance.options.except(:klass)
       end
 
+      # Added as a field to ensure copying this value to any blocks created
+      # from a draft.
+      field :activate_with, type: String
```
bencrouse: I believe this didn't come over in the code above looping over fields because it's not actually a Mongoid field, it's just an `attr_accessor` in `Releasable`. Can you update to match?

mttdffy: No. I did this specifically because we want the attr_accessor on a block to be set if its created from a draft, which making it a field on the draft will accomplish...thus the comment above it.

bencrouse: Ok, I'm thinking blocks only get instantiated for previewing. But I'm not sure it's an important difference, this may be useful down the line.

PR #264, opened 2019-12-02
Restrict release datetimepicker to dates in the future

mttdffy:
WORKAREA-65

### DIFF
--- a/admin/app/views/workarea/admin/create_releases/setup.html.haml
+++ b/admin/app/views/workarea/admin/create_releases/setup.html.haml
@@ -22,12 +22,12 @@
             .grid__cell.grid__cell--50-at-medium
               .property
                 = label_tag 'release_publish_at', t('workarea.admin.fields.publish_at'), class: 'property__name'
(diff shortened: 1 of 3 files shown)


mttdffy:
> Does this use of `minDate` still allow a past time? e.g. if you select today, is the time selected midnight and therefore actually in the past? Might need to make it a validation on the model.

We do: https://github.com/workarea-commerce/workarea/blob/v3.4-stable/core/app/models/workarea/release.rb#L198-L202

Except It doesn't validate this after its been published. So I guess we _could_ change it to not check `#published?`, but im not sure the reasoning behind why we did in the first place

bencrouse:
I'm guessing it's because of allowing potential changes (or necessary changes) to the `Release` after it's been published and not having those fail validation.

It seems fine to leave it just as a client-side thing, but does `minDate` of `0` mean you can pick today and then today's midnight is in the past?

bencrouse:
I guess I'm asking: does `minDate` of `1` mean tomorrow which seems like the better restriction?

mttdffy:
So they couldn't schedule a release to be the day they create it?

bencrouse:
Oh my bad - forgot it was also a time picker (not just date).

PR #267, opened 2019-12-03
Improve presentation of nested lists

chriscressman:
Nested lists look awkward because whitespace is applied only to the
bottom of each list item, so there is no separation between the raw text
within a list item and the list that follows it.

Move list item margin from the bottom to the top to fix this issue.
There are no unwanted side effects at the top or bottom of each list,
because each ul/ol has its own top and bottom margin which is greater
than the margin applied to the li.

WORKAREA-127


chriscressman:
Before:

before

After:

after

PR #268, opened 2019-12-03
Fix/improve home page copy

chriscressman:
On the docs home page, "publically" is a misspelling, and "bi-monthly"
is used incorrectly ("bi-weekly" or "semi-monthly" was likely intended).

Fix the spelling error and rewrite the email signup copy.

WORKAREA-129


chriscressman:
Before:

before

After:

after

PR #270, opened 2019-12-04
Fix password config not available when building indexes

bencrouse:
This causes a null value for expireAfterSeconds when creating indexes in
Mongo.


bencrouse:
https://github.com/workarea-commerce/workarea/blob/master/core/app/models/workarea/user/password_reset.rb#L11

PR #272, opened 2019-12-04
Improve documentation feedback process

chriscressman:
Documentation feedback is collected anonymously through a Google form.
This does not allow for communication with the reporter, and it is an
additional communication channel that has to be managed separately.

Use GitHub issues instead of the Google form to collect documentation
feedback. This allows triaging docs issues using the same process used
for the software and allows communication with the reporter (to thank
them, ask questions, provide a status, or help them if the issue is not
actually a documentation problem).

Also update the documentation issue template to broaden its scope and
provide a default assignee.

WORKAREA-64

No changelog

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation-request.md
+++ b/.github/ISSUE_TEMPLATE/documentation-request.md
@@ -1,9 +1,9 @@
 ---
-name: Documentation request
-about: Suggest documentation
+name: Documentation feedback/request
+about: Provide documentation feedback
 title: ''
 labels: documentation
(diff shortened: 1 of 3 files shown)


chriscressman:
New doc issue "landing" **before**:

issue-github-before

New doc issue "landing" **after**:

issue-github-after

Doc UI **before**:

docs-before

Doc UI **after**:

docs-after

New doc issue "landing" when created from doc:

docs-issue

bencrouse:
Nice

PR #275, opened 2019-12-05
Exclude Update Timestamp From Imports

tubbo:
Update the `:updated_at` timestamp to the current time when existing
models are updated via an import, and ignore any settings of the
`:updated_at` timestamp in JSON/CSV imports, as this can interfere with
cache key generation.

### DIFF
--- a/core/app/models/workarea/data_file/csv.rb
+++ b/core/app/models/workarea/data_file/csv.rb
@@ -48,6 +48,8 @@ module Workarea
 
       def assign_attributes(model, attrs)
         model.fields.each do |name, metadata|
+          next if name == 'updated_at'
+
           value = CsvFields.deserialize_from(attrs, field: metadata, model: model)
           model.send("#{name}=", value) if value.present?
         end
--- a/core/app/models/workarea/data_file/json.rb
+++ b/core/app/models/workarea/data_file/json.rb
@@ -45,10 +45,29 @@ module Workarea
 
         if id.present?
           result = model_class.find_or_initialize_by(id: id)
-          result.attributes = attributes
+          result.attributes = attributes_without_updated_at(attributes)
           result
         else
           model_class.new(attributes)
+        end
+      end
+
+      def attributes_without_updated_at(attrs)
+        return attrs unless attrs.respond_to?(:each_with_object)
+
+        attrs.each_with_object({}) do |(key, value), attributes|
+          next if key.to_s == 'updated_at'
+
+          attributes[key] = case value
+                            when Hash
+                              attributes_without_updated_at(value)
+                            when Array
+                              value.map do |item|
+                                attributes_without_updated_at(item)
+                              end
+                            else
+                              value
+                            end
         end
       end
     end
--- a/core/test/models/workarea/data_file/csv_test.rb
+++ b/core/test/models/workarea/data_file/csv_test.rb
@@ -450,6 +450,22 @@ module Workarea
         assert_equal(Qoo, model.bars.second.class)
         assert_equal('2', model.bars.second.qoo)
       end
+
+      def test_exclude_updated_at
+        original_date = 2.days.ago
+        model = Foo.create!(name: '1', updated_at: original_date)
+        model.name = '2'
+        csv = Csv.new.serialize(model)
+        import = create_import(
+          model_type: Foo.name,
+          file: create_tempfile(csv, extension: 'csv'),
+          file_type: 'csv'
+        )
+
+        assert_changes -> { model.reload.updated_at.to_date } do
+          Csv.new(import).import!
+        end
+      end
     end
   end
 end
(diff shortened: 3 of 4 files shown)


core/app/models/workarea/data_file/json.rb
```diff
@@ -45,12 +45,55 @@ def find_updated_model_for(attrs)
 
         if id.present?
           result = model_class.find_or_initialize_by(id: id)
-          result.attributes = attributes
+          result.attributes = attributes_without_updated_at(attributes)
+          result.touch if result.changed?
```
tubbo: I'm not sure if this is a Mongoid bug, but the reason I had to go ham on the recursion with `#touch` was because when you mass-assign attributes like this it actually doesn't seem to touch the embedded models...

core/app/models/workarea/data_file/csv.rb
```diff
@@ -48,12 +48,22 @@ def export!
 
       def assign_attributes(model, attrs)
         model.fields.each do |name, metadata|
+          next if name == 'updated_at'
+
           value = CsvFields.deserialize_from(attrs, field: metadata, model: model)
-          model.send("#{name}=", value) if value.present?
+          changed = if value.is_a?(Time)
```
bencrouse: Do you need this if we're just skipping `updated_at`?

core/app/models/workarea/data_file/csv.rb
```diff
@@ -48,12 +48,22 @@ def export!
...
+                    end
+
+          model.send("#{name}=", value) if value.present? && changed
         end
 
-        model.updated_at ||= Time.current
-        model.created_at ||= Time.current
+        if model.embedded? && model.changed?
+          model.updated_at = Time.current
+          model.created_at ||= Time.current
```
bencrouse: Putting this in an `if` is a change, we'll always want to ensure it.

core/app/models/workarea/data_file/json.rb
bencrouse: `touch` triggers callbacks and db persistence iirc, we won't want to do that.

core/app/models/workarea/data_file/csv.rb
tubbo: I was seeing `updated_at` getting changed on records that didn't actually get updated, so for example if you exported a CSV, changed one line, and re-imported it, _all_ of the entries would get updated, which would break a whole lot of cache entries unnecessarily.

core/app/models/workarea/data_file/csv.rb
tubbo: This fixes a bug with `created_at` always causing the record to update even though nothing changed. The CSV imports had an underlying bug wherein any records that are imported are touched, even though they don't actually get updated.

bencrouse: I'm confused - below in the json version, you're specifically writing code to touch everything. Why can't we just let the oob Mongoid behavior do it's thing?

core/app/models/workarea/data_file/csv.rb
bencrouse: I mean specifically the `created_at` should not be an `if`, we always want to ensure that's present.

core/test/models/workarea/data_file/csv_test.rb
```diff
@@ -450,6 +450,77 @@ def test_embedded_subclasses
...
+        import = create_import(
+          model_type: Foo.name,
+          file: create_tempfile(csv, extension: 'csv'),
+          file_type: 'csv'
+        )
+
+        Csv.new(import).import!
+
+        assert_equal('2', model.reload.name)
+        assert_equal(Time.current.to_date, model.updated_at.to_date)
```
bencrouse: Why compare dates here?

core/test/models/workarea/data_file/csv_test.rb
```diff
@@ -450,6 +450,77 @@ def test_embedded_subclasses
...
+        csv = Csv.new.serialize(model)
+        import = create_import(
+          model_type: Foo.name,
+          file: create_tempfile(csv, extension: 'csv'),
+          file_type: 'csv'
+        )
+
+        Csv.new(import).import!
+
+        assert_equal(original_date.to_date, model.reload.updated_at.to_date)
```
bencrouse: Dates don't feel like the best solution here - maybe try an `assert_changes`?

core/app/models/workarea/data_file/csv.rb
tubbo: I tried that first and it didn't work. When adding the tests for what happens in an embedded record, I was not seeing updated_at changing.

core/app/models/workarea/data_file/csv.rb
tubbo: wouldn't mongoid do that for us?

bencrouse: If you pass an invalid `created_at`, Mongoid will leave it as `nil` iirc :(

PR #280, opened 2019-12-09
Fix/improve section linking

chriscressman:
Fix/improve section identification & navigation

The migration from the previous documentation toolchain to the current
toolchain introduced several problems regarding document sections:

* Section IDs changed, leaving most links to sections broken.
* Clicking the "link" icon next to a heading loads a broken URL when the
  current URL already contains a fragment ID.
* When loading a specific section, the section's header is not visible
  and requires scrolling up to see it.
* Clicking a link in the in-page table of contents scrolls to the
  section, but does not update the current URL or browser history stack.
* When the URL contains a fragment ID, the corresponding section is not
  highlighted in any way.
* The logic used to generate IDs for sections is fragile, resulting in
  section IDs changing when headings are added or changed in unrelated
  parts of the documents. This leads to authors breaking links to
  sections without realizing it.

Therefore, apply the following fixes and improvements:

* Restore the previous logic for generating section IDs, fixing links
  to sections migrated from the previous toolchain and making links to
  sections reliable moving forward.
* Fix the href of the "link" icon when a fragment ID is already present.
* Use the default browser behavior for links within the table of
  contents, which ensures the current URL, browser history, and scroll
  position all work as expected.
* Apply styles that fix the scroll position when a fragment ID is
  present and highlight the selected section (by showing its "link"
  icon).

---

Fix href and src values

Restoring the old section ID logic fixed the majority of broken links to
sections, but it broke all links to sections that were written _after_
the migration to the current toolchain.

Fix those href values to use the "old" values to make all well, now and
moving forward.

While doing this, also fix href and src values in general. The current
toolchain requires all href and src values to be root-relative (begin
with "/") for them to be portable (continue working if the article in
which they appear is moved).

WORKAREA-63

No changelog


chriscressman:
I committed the changes to the build system and the docs separately to make it easier to review the actual code changes.

PR #282, opened 2019-12-11
Improve Redis configuration defaulting

bencrouse:
This makes Redis configuration a little more robust, allowing partial
configuration values that will always end up falling back to defaults.

### DIFF
--- a/core/lib/workarea/configuration/redis.rb
+++ b/core/lib/workarea/configuration/redis.rb
@@ -34,41 +34,37 @@ module Workarea
         def find_config(name)
           config_slug = name.to_s.underscore.downcase
           from_config = Workarea.config[config_slug].presence ||
-                          Rails.application.secrets[config_slug]
-
-          return from_config if from_config.present?
+                          Rails.application.secrets[config_slug] ||
+                          {}
 
           env_slug = name.to_s.underscore.upcase
-          scheme = ENV["WORKAREA_#{env_slug}_SCHEME"].presence || DEFAULT[:scheme]
-
-          {
-            scheme: scheme,
-            host: ENV["WORKAREA_#{env_slug}_HOST"].presence || DEFAULT[:host],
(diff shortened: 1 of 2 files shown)


core/lib/workarea/configuration/redis.rb
```diff
@@ -34,21 +34,17 @@ def cache
...
-            password: ENV["WORKAREA_#{env_slug}_PASSWORD"].presence,
-            ssl: scheme == 'rediss' ? true : false
-          }
+          from_config.symbolize_keys.reverse_merge(
+            scheme: ENV["WORKAREA_#{env_slug}_SCHEME"],
+            host: ENV["WORKAREA_#{env_slug}_HOST"],
+            port: ENV["WORKAREA_#{env_slug}_PORT"],
+            db: ENV["WORKAREA_#{env_slug}_DB"],
+            password: ENV["WORKAREA_#{env_slug}_PASSWORD"]
+          )
```
MMartyn: Looks like ssl key was removed here.

bencrouse: Yeah that was intentional, it seemed to make more sense as a method if it isn't configurable or anything. Is `find_config` being used somewhere where that would matter?

bencrouse: How/where was that getting used in the first place?

MMartyn: It is used in the redis client. See:
https://github.com/redis/redis-rb/blob/master/lib/redis/client.rb#L441
https://github.com/workarea-commerce/workarea/blob/e46d7ef52901aab16fc7d683aa319cd68c71dcd5/core/lib/workarea.rb#L50
https://github.com/workarea-commerce/workarea/blob/master/core/lib/workarea/configuration/redis.rb#L56

bencrouse: Ah my bad. Fix impending.

PR #284, opened 2019-12-12
Slightly more specific language

leviathant:
Based on feedback from Lou Perseghin, I wanted to make it clear that in order to install Workarea locally, you need to install Docker, and that the command you're running is being run in terminal.


bencrouse:
Thanks @leviathant !

PR #285, opened 2019-12-12
Disallow multiple form submissions throughout the Admin

meowsus:
WORKAREA-133


meowsus:
> In the storefront, we use `disable_with` from the UJS, any reason we can't use that here?

@bencrouse I didn't want to edit, like, every view in a patch and I'm 99% sure it wouldn't work with Turbolinks because Turbolinks would cache the disabled state of the button. Returning to the previous page would show the button as disabled still.

bencrouse:
Ok, I think there needs to be some visual representation of what's happening for UX reasons. If something goes wrong (e.g. something with Turbolinks), we don't want a button that looks like it's all set for clicking, but actually does nothing.

Can we do something where we're not reimplementing UJS then?

meowsus:
I'm stupid. Turbolinks caching doesn't apply to form submissions, unless we're using jQuery UJS, but that's a whole separate issue. If we are using jQuery UJS for a form submission, and the problem this PR is solving is an issue with that specific interaction, we can just use `disable-with` in those specific cases.

bencrouse:
How about something like this to use UJS instead of writing our own:

```
def button_tag(content_or_options = nil, options = nil, &block)
  options ||= {}
  options[:data] ||= {}
  options[:data][:disable_with] ||= t('blah')
  super(content_or_options, options, &block)
end
```

I guess I'm feeling like we should be leveraging the frameworks we have instead of writing our own code.

bencrouse:
I'm willing to take a stab at getting a helper override version to work if you're trying to move on to other things.

tubbo:
> I guess I'm feeling like we should be leveraging the frameworks we have instead of writing our own code.

If we're going to do this, maybe it's time to look at `form_with` (fun fact: it's `:remote` by default!) and form builders instead of copypasting hundreds of lines of Haml every time you want to make an admin view?

meowsus:
@bencrouse I mean if you feel strongly about it, go right ahead. I think that this module is pretty low impact and solves the problem we're trying to solve without having to monkey patch or edit a billion views.

bencrouse:
I don't want to maintain code we already have a library for. I'll see what I can do.

meowsus:
@bencrouse you wants I should merge this?

bencrouse:
I gotchu.

PR #296, opened 2019-12-19
Use the Rack session ID cookie value for user activity session IDs

bencrouse:
Rack >= 2.0.8 adds the idea private/public session IDs to prevent timing
attacks where a session ID can be stolen. This is big for sessions stored
in databases because the session can then be stolen.

Workarea only supports a cookie session store, so we can continue to
safely use the cookie value of the session ID for metrics lookups.

You can learn more about the Rack vulnerability here:
https://github.com/rack/rack/security/advisories/GHSA-hrqr-hxpp-chr3


bencrouse:
Not sure why this is open, I fixed this shit on Weblinc day.

PR #306, opened 2020-01-06
Add product tags to storefront search index

mttdffy:


core/app/models/workarea/search/storefront/product.rb
```diff
@@ -59,7 +59,8 @@ def content
             category_names: category_names,
             description: catalog_content,
             details: details_content,
-            facets: facets_content
+            facets: facets_content,
+            tags: model.tag_list
```
bencrouse: I think this should be in `keywords` or maybe `facets` - `content` gets fulltext-analyzed so we def wouldn't tags to get that analysis for this purpose.

core/app/models/workarea/search/storefront/product.rb
```diff
@@ -38,7 +38,8 @@ def keywords
           {
             catalog_id: clean_for_keywords(catalog_id),
             sku: sku.map { |s| clean_for_keywords(s) },
-            name: clean_for_keywords(model.name)
+            name: clean_for_keywords(model.name),
+            tags: model.tag_list
```
bencrouse: Seems like this should just be an array so we can provide exact matching without worrying about commas.

PR #307, opened 2020-01-06
Another hardcoded 2020 fix

bencrouse:
We've all learned our lesson, right?

### DIFF
--- a/testing/lib/workarea/storefront/integration_test.rb
+++ b/testing/lib/workarea/storefront/integration_test.rb
@@ -66,7 +66,7 @@ module Workarea
             credit_card: {
               number: '1',
               month:  1,
-              year:   2020,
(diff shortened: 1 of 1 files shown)


testing/lib/workarea/storefront/integration_test.rb
```diff
@@ -66,7 +66,7 @@ def complete_checkout(email = nil, password = nil)
             credit_card: {
               number: '1',
               month:  1,
-              year:   2020,
+              year:   next_year,
```
meowsus: Why not `twenty_twenty`? :trollface:

bencrouse: `alias_method twenty_twenty next_year`

accepting PRs...

PR #310, opened 2020-01-08
Remove wait_for_iframe

mcfearsome:
synchronize sleeps in 0.01 increments up to the default_wait_time of Capybara.  The find_button triggers its own synchronize but I think there is a bug that causes a loss of context of the within_frame, the outer synchronize seems to solve this


bencrouse:
Adding @mttdffy and @meowsus because they also spent some time looking into this.

mcfearsome:
Sure thing @bencrouse, I'll clean it up today or tomorrow

PR #312, opened 2020-01-10
Link to first "footer" area in layout Content from Shortcuts

meowsus:
We need to be more permissive in our linking to footer content areas
from the header, since themes and builds can technically rename these
areas. Now this link will point to the first content area that contains
the word 'footer'.

WORKAREA-145

### DIFF
--- a/admin/app/helpers/workarea/admin/content_helper.rb
+++ b/admin/app/helpers/workarea/admin/content_helper.rb
@@ -6,6 +6,16 @@ module Workarea
         @tmp[name.to_s.systemize] ||= Content.for(name.to_s.systemize)
       end
 
+      def layout_footer_area_id
+        @layout_footer_area_id ||= begin
+          content = Admin::ContentViewModel.wrap(
+            current_system_page_content_for(:layout)
+          )
+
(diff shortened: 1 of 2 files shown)


admin/app/helpers/workarea/admin/content_helper.rb
```diff
@@ -6,6 +6,15 @@ def current_system_page_content_for(name)
         @tmp[name.to_s.systemize] ||= Content.for(name.to_s.systemize)
       end
 
+      def edit_layout_footer_area_path
+        content = Admin::ContentViewModel.wrap(
+          current_system_page_content_for(:layout)
+        )
+
+        area_id = content.areas.find { |a| a =~ /footer/i }
+        edit_content_path(content, area_id: area_id || content.areas.first)
```
bencrouse: I think it would be better to just not render the link if we can't find an area explicitly named with "footer". Like I could see this logic end up linking to "header promo" area. That would be a weird UX then.

meowsus: Yep, that makes more sense.

PR #321, opened 2020-01-21
Fix nil segment IDs

bencrouse:
This will fix errors raised when `active_segment_ids` are `nil`.


bencrouse:
Seen here: https://github.com/workarea-commerce/workarea-b2b/runs/399741966#step:4:494

PR #325, opened 2020-01-22
Overhaul "testing" docs

chriscressman:
WORKAREA-150


docs/source/articles/testing-concepts.html.md
```diff
@@ -0,0 +1,116 @@
...
+
+
+## Test Case Types
+
+Workarea defines its own _test case types_, and all Workarea test cases inherit from one of those types.
+
+When [writing your own tests](/articles/decorate-and-write-tests.html), choose the appropriate test case type.
+Refer to the following table for brief descriptions, and click through to the sources to see the full implementation of each type.
+
+| Test Case Type | Description |
```
chriscressman: I didn't go as far as originally planned this this (i.e. writing full descriptions in RDoc) because I ran out of time, but moving these into a table with source code links should make it easier to separate out this reference material in the future if you decide to do so.

docs/source/articles/configure-headless-chrome.html.md
```diff
@@ -0,0 +1,26 @@
...
+
+| Config | Description |
+| --- | --- |
+| [`headless_chrome_options`](https://github.com/workarea-commerce/workarea/blob/v3.5.3/core/lib/workarea/configuration.rb#L953) | Options hash passed to the Selenium driver |
+| [`headless_chrome_args`](https://github.com/workarea-commerce/workarea/blob/v3.5.3/core/lib/workarea/configuration.rb#L956-L962) | Array of command line arguments passed to the Chrome executable |
+
+Replace or mutate the default values from within your own initializer:
+
+```
+# <your_application_root>/config/initializers/headless_chrome.rb
```
bencrouse: Feels like this should be in `test/test_helper.rb` or something like `test/support/headless_chrome.rb` to indicate this exists for testing purposes only.

docs/source/articles/decorate-and-write-tests.html.md
```diff
@@ -0,0 +1,218 @@
...
+This doc explains how to [Decorate a Workarea Test Case](#decorate-a-workarea-test-case), including the special case [Skip a Workarea Test](#skip-a-workarea-test).
+
+This doc also covers how to [Write an Application Test Case](#write-an-application-test-case), and when doing so, the following recipes may also be useful:
+
+* [Change Configuration within a Test](#change-configuration-within-a-test)
+* [Change Locale within a Test](#change-locale-within-a-test)
+* [Change Time within a Test](#change-time-within-a-test)
+
+
+## Decorate a Workarea Test Case
```
bencrouse: There's one point of guidance not mentioned here, I'm not sure exactly what the best explanation would be. The idea is that decorating tests to add tests is almost always better done in a new test file, like `test/models/workarea/payment/my_site_credit_card_integration_test.rb` It makes it way easier to write the new test in terms of running and traces. It also gives you the ability to do only the test setup you need for your actual tests instead of the setup built for the ones in the base test class.

docs/source/articles/run-tests.html.md
```diff
@@ -0,0 +1,203 @@
+---
+title: Run Tests
```
bencrouse: I'm not sure where this should be, but it would be good to offer suggestions about setting up builds which is mostly running tests. Like something about parallelizing core/admin/storefront/plugins or some such to improve build times.

docs/source/articles/testing-concepts.html.md
```diff
@@ -0,0 +1,116 @@
...
+Refer to that document for Rails testing fundamentals, such as Minitest, test cases, and assertions.
+
+However, a new Workarea application includes a broad feature set by default and therefore includes _tests_ for those features.
+You can run the platform's own tests (referred to here as _Workarea tests_, to differentiate them from application tests) from your application.
+( See [Run Tests](/articles/run-tests.html). )
+
+You can also extend the platform's test suite by writing _decorators_ for existing tests and writing your own tests (referred to here as _application tests_ to differentiate them from Workarea tests).
+( See [Decorate & Write Tests](/articles/decorate-and-write-tests.html). )
+
+When you extend Workarea with your own customizations and functionality, you should extend the test suite as well.
```
bencrouse: 👍 

I think this could probably be mentioned in the decoration article too.

docs/source/articles/decorate-and-write-tests.html.md
jyucis: Ben makes a good point. The guidance I usually give around this is that a developer should only decorate a test if the host app breaks or invalidates the core test. Otherwise it is best practice to create new test files.

docs/source/articles/extend-test-case-mixins.html.md
```diff
@@ -0,0 +1,61 @@
...
+* [Add a New Test Case Mixin](#add-a-new-test-case-mixin)
+
+
+## Change an Existing Test Case Mixin
+
+You may want to change a method within an existing test case mixin so that the change is reflected in all tests (rather than decorating specific tests).
+
+For example, say you want to change the email address filled in by [`Storefront::SystemTest#fill_in_email`](https://github.com/workarea-commerce/workarea/blob/v3.5.3/testing/lib/workarea/storefront/system_test.rb#L90-L92) from `bcrouse-new-account@workarea.com` to `robert-clams@workarea.com`.
+
+Create a new file in `/test/support/` and re-open the previously defined module and method, applying the email address change:
```
jyucis: I think its worth mentioning that any files in test/support will be loaded when the tests are run.

docs/source/articles/run-tests.html.md
chriscressman: I recommend writing a how to doc for setting up CI, and reference the testing docs as necessary from there.

chriscressman:
Addressed feedback in PR and also updated time travel example, which we discussed in our meeting.

PR #327, opened 2020-01-23
Fix docs building

bencrouse:


bencrouse:
Any thoughts on committing the `Gemfile.lock` so we don't hit similar in the future? Do we really care about using the latest gems to build static HTML?

chriscressman:
> Any thoughts on committing the `Gemfile.lock` so we don't hit similar in the future? Do we really care about using the latest gems to build static HTML?

Gemfile.lock used to be in the repo (when the source was in Stash). I explicitly committed it because the docs are an application, not a library, so everyone running it (including nonhumans) should be running the same version of everything.

PR #328, opened 2020-01-23
Allow custom naming of Content Blocks

meowsus:
WORKAREA-171

### DIFF
--- a/admin/app/views/workarea/admin/content/_form.html.haml
+++ b/admin/app/views/workarea/admin/content/_form.html.haml
@@ -15,9 +15,18 @@
     .grid__cell
       = content_block_icon(block.icon, title: block.name, class: 'svg-icon svg-icon--gray')
     .grid__cell
-      %span.content-editor__form-heading= block.name
+      .grid.grid--middle.grid--auto
+        .grid__cell
+          %span.content-editor__form-heading= block.name
+        .grid__cell
+          = link_to inline_svg_tag('workarea/admin/icons/edit.svg', class: 'svg-icon svg-icon--small svg-icon--dark-gray', title: t('workarea.admin.content.form.edit_block_name')), '#', data: { toggle_element: "##{dom_id(block, 'content_block_name_field')}" }
 
 .content-editor__form-fields
+  .hidden{ id: dom_id(block, 'content_block_name_field') }
+    .property
+      = label_tag 'block_name', t('workarea.admin.content.form.new_block_name'), class: 'property__name align-center'
+      = text_field_tag 'block[name]', block.name, id: nil, class: 'text-box text-box--i18n'
+
   .tabs
     - block.fieldsets.each do |fieldset|
       .tabs__panel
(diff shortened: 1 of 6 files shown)


meowsus:
Screenshot from 2020-01-23 14-38-34

admin/app/views/workarea/admin/content/_form.html.haml
```diff
@@ -11,13 +11,11 @@
...
       = link_to "← #{t('workarea.admin.content.form.all_blocks')}", '#', data: { content_editor_form_cancel: '' }
 
-  .grid.grid--middle.grid--center.grid--auto
-    .grid__cell
-      = content_block_icon(block.icon, title: block.name, class: 'svg-icon svg-icon--gray')
-    .grid__cell
-      %span.content-editor__form-heading= block.name
-
 .content-editor__form-fields
+  .property
```
tubbo: are we losing the icon here? I feel like it might help to keep it around..

bencrouse:
At the UX meeting we talked about this being a toggling field from the main list so as not to clutter things up on the already squished edit form. Did you explore that?

meowsus:
> At the UX meeting we talked about this being a toggling field from the main list so as not to clutter things up on the already squished edit form. Did you explore that?

@bencrouse:

I did, and decided on this simpler approach.

When viewing the list of content blocks we would have had 3 interactions on each row: click to drag, click to edit, and click to expand. Click to drag and to expand actions (what we have currently) play well together currently because the the drag action and expand action triggers are separated in the the UI.

Putting an edit icon on that page means that a non-trivial part of that module would have to be rewritten just to support this new action, since it would exist within the context of the expand action. Assuming that the updating of the block's name should be async, since a full page refresh would be annoying due to the load behavior of this rather large page, would be annoying. Adding error states in this case translates to more non-trivial updates.

Realizing that the simplest solution also required writing less code and accounting for far fewer use cases, I went with this solution that I've implemented. The "more Railsy" solution.

Since we're speaking subjectively, I don't feel like the edit form is "squished" but perhaps I'd agree that it feels "more cluttered." I'd be happy with a compromise where:

The user:

1. clicks to expand the block edit form
1. sees the icon & name (as previously displayed) supplemented by an edit icon
1. clicks the icon
1. sees the name exchanged for a text box

How's them apples?

core/app/models/workarea/content/block.rb
```diff
@@ -39,6 +39,11 @@ class Block
       #
       field :hidden_breakpoints, type: Array, default: []
 
+      # @!attribute custom_name
+      #   @return [String] the user-defined name of the block
+      #
+      field :custom_name, type: String, default: nil
```
mttdffy: default is always nil if not defined, no need to be explicit. Think the field should just be `name` and then you can define the getter

```
def name
  super || BlockName.new(self).to_s
end
```

meowsus:
Before click:
Screenshot from 2020-02-06 15-02-41

After click:
Screenshot from 2020-02-06 15-03-01

PR #331, opened 2020-01-24
Fix order of steps to create new app

chriscressman:
The procedure in "Create a New Workarea App" fails at step 3:
"Install Workarea into the Rails application", because that step
requires MongoDB to be running.

Fix the procedure by transposing steps 3 and 4, resulting in
"Start Workarea service dependencies" running before "Install Workarea
into the Rails application", which ensures MongoDB is available.

I re-tested the entire procedure from scratch to ensure this works.

WORKAREA-186


bencrouse:
Fixes #337

PR #336, opened 2020-01-24
Add image dimensions to Asset summaries

meowsus:
WORKAREA-173


admin/app/views/workarea/admin/content_assets/_summary.html.haml
```diff
@@ -2,7 +2,10 @@
   .summary__image-container
     - if model.image?
       = image_tag model.file.process(:medium).url, alt: model.name, class: 'summary__image', title: model.alt_text
-  %span.summary__name= model.name
+  %span.summary__name<
+    = model.name
+    %br
+    #{model.file_width}x#{model.file_height}
```
bencrouse: Since this partial is also used in search results, I think we should only output this if the `model.image?`

PR #340, opened 2020-01-27
Fix Teaspoon in CI

meowsus:
This was no longer running properly due to a strange `coffee_rails`
dependency. More information here:

https://github.com/jejacks0n/teaspoon/issues/405

WORKAREA-185

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,4 +10,5 @@ gem 'listen'
 gem 'bundler-audit'
 gem 'rubocop'
 gem 'rails-decorators', git: 'https://github.com/workarea-commerce/rails-decorators.git'
-gem 'teaspoon', git: 'https://github.com/jtapia/teaspoon.git', branch: 'chore/update-rails-6'
+
+gem 'teaspoon', git: 'https://github.com/jejacks0n/teaspoon.git'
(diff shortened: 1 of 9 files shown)


Gemfile
```diff
@@ -10,4 +10,6 @@ gem 'listen'
 gem 'bundler-audit'
 gem 'rubocop'
 gem 'rails-decorators', git: 'https://github.com/workarea-commerce/rails-decorators.git'
-gem 'teaspoon', git: 'https://github.com/jtapia/teaspoon.git', branch: 'chore/update-rails-6'
+
+gem 'teaspoon', git: 'https://github.com/jejacks0n/teaspoon.git'
+gem 'coffee-script', group: :test
```
bencrouse: From that issue, looks like it was teaspoon fixed years ago, things run fine on my local copy without this. What was the weirdness?

meowsus: My local was hanging when running the real test suite. I found the dependency after running `teaspoon:server`. It was printing out a unmet dependency error around coffee-script. This cleared it up for me, but I'll re-check without it.

meowsus: Dunno what was going on in my brain

PR #343, opened 2020-01-30
Add ability to disable audit log on a model

mttdffy:
WORKAREA-167

I pointed this at v3.5-stable because it could easily be a patch, but we can change it we want. This would less awkward if we just patched mongoid-audit_log

### DIFF
--- a/core/app/models/workarea/audit_logging.rb
+++ b/core/app/models/workarea/audit_logging.rb
@@ -1,0 +1,32 @@
+module Workarea
+  module AuditLogging
+    extend ActiveSupport::Concern
+    include Mongoid::AuditLog
+
+    class_methods do
+      def enable_audit_log
+        @audit_log_enabled = true
+      end
+
+      def disable_audit_log
+        @audit_log_enabled = false
+      end
+
+      def audit_log_enabled?
+        !defined?(@audit_log_enabled) || @audit_log_enabled
+      end
+    end
+
+    private
+
+    def set_audit_log_changes
+      return unless self.class.audit_log_enabled?
+      super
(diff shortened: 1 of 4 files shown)


bencrouse:
Yeah, changes look good, but why don't we just do this in the the audit log gem?

mttdffy:
> Yeah, changes look good, but why don't we just do this in the the audit log gem?

No reason, thats why I said that in the top comment :) 

It was never mentioned when we talked about it that we do it there, so the rationale _could_ be that in a normal scenario you are explicitly including audit log in each model so disabling after including the module seems like getting in the pool but wanting to stay dry. So in that case, one could make the argument that this isn't really necessary in a world where you don't have it being included globally via an `ApplicationDocument` like Workarea does and thusly its a Workarea-only problem.

That said, I don't think there is any harm in adding it to the gem. So I will do that.

core/app/models/workarea/audit_logging.rb
```diff
@@ -0,0 +1,32 @@
+module Workarea
+  module AuditLogging
+    extend ActiveSupport::Concern
+    include Mongoid::AuditLog
+
+    class_methods do
+      def enable_audit_log
+        @audit_log_enabled = true
```
eric-pigeon: Should these be thread scoped?

mttdffy: no, looking to disable permanently.

mttdffy: We could probably do something a bit more useful in the gem itself with allowing temporary disabling/enabling as an option

PR #344, opened 2020-01-31
Allow images to be lazy loaded

meowsus:
WORKAREA-194

### DIFF
--- a/storefront/app/assets/stylesheets/workarea/storefront/base/_images.scss
+++ b/storefront/app/assets/stylesheets/workarea/storefront/base/_images.scss
@@ -27,3 +27,15 @@ img {
 img[width], img[height] { /* [2] */
     max-width: none;
 }
+
+
+/**
+ * Styles for images while they're lazy loading. Classes supplied via JS.
+ */
+img.lazy-image {
+    opacity: 0;
+}
+
+img.lazy-image--loaded {
+    opacity: 1;
+}
--- a/storefront/app/helpers/workarea/storefront/application_helper.rb
+++ b/storefront/app/helpers/workarea/storefront/application_helper.rb
@@ -90,6 +90,35 @@ module Workarea
         content_tag(:div, capture(&block), { class: 'hidden-if-js-enabled',
                                              data: { optional_field: prompt } })
       end
+
+      def lazy_image_tag(source, options = {})
+        options = options.symbolize_keys
+
+        check_for_image_tag_errors(options)
+        skip_pipeline = options.delete(:skip_pipeline)
+
+        if options[:srcset] && !options[:srcset].is_a?(String)
+          options[:srcset] = options[:srcset].map do |src_path, size|
+            src_path = path_to_image(src_path, skip_pipeline: skip_pipeline)
+            "#{src_path} #{size}"
+          end.join(", ")
+        end
+
+        options[:width], options[:height] = extract_dimensions(options.delete(:size)) if options[:size]
+
+        lazy_options = options.deep_dup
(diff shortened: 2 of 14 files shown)


storefront/app/assets/stylesheets/workarea/storefront/base/_images.scss
```diff
@@ -27,3 +27,15 @@ img {
 img[width], img[height] { /* [2] */
     max-width: none;
 }
+
+
+/**
+ * Styles for images while they're lazy loading
+ */
+img.lazy-image {
```
bencrouse: Is this necessarily a good idea? Last time I read up, using progressive JPGs was good to allow them to see that images are coming in (also why our jpeg processing converts to progressive).

storefront/app/helpers/workarea/storefront/application_helper.rb
```diff
@@ -90,6 +90,32 @@ def optional_field(prompt, *fields, &block)
         content_tag(:div, capture(&block), { class: 'hidden-if-js-enabled',
                                              data: { optional_field: prompt } })
       end
+
+      def image_tag(source, options = {})
+        options = options.symbolize_keys
+        check_for_image_tag_errors(options)
+        skip_pipeline = options.delete(:skip_pipeline)
+        options[:lazy_load] = options[:lazy_load].presence || true
```
bencrouse: Unfortunately, 
```
irb(main):001:0> false.present?
=> false
```

So this should be something like `options.key?(:lazy_load) ? options[:lazy_load] : true`


In general for this method though, do we need to copy the implementation from Rails in? Could it be more like:


```
def image_tag(source, options = {})
  options = options.deep_dup
  if options.fetch(:lazy_load, true)
    options[:data] ||= {}
    options[:data][:lazy_image] = resolve_image_source(source, options[:skip_pipeline])
    options[:class] ||= ''
    options[:class] << ' lazy-image'
    source = nil
  end

  super(source, options)
end
```

bencrouse:
We used to have an olde `noscript` in there for no-JS fallback, do we care about that now @meowsus ?

storefront/app/helpers/workarea/storefront/application_helper.rb
meowsus: Unfortunately `source` can't be `nil`. You'll get a `Nil location provided. Can't build URI.` error. I'll mess with it.

meowsus:
Alright; after much discourse @bencrouse decided to go back to the `lazy_image_tag` helper because there was no reasonable way to trick a monkey-patched `image_helper` into completely removing the `src` attribute from the resulting markup.

storefront/app/helpers/workarea/storefront/application_helper.rb
```diff
@@ -90,6 +90,39 @@ def optional_field(prompt, *fields, &block)
...
+          options[:srcset] = options[:srcset].map do |src_path, size|
+            src_path = path_to_image(src_path, skip_pipeline: skip_pipeline)
+            "#{src_path} #{size}"
+          end.join(", ")
+        end
+
+        options[:width], options[:height] = extract_dimensions(options.delete(:size)) if options[:size]
+
+        image_source = resolve_image_source(source, skip_pipeline)
+        if options.fetch(:lazy_load, true)
```
bencrouse: Wouldn't this always be true since you're calling `lazy_image_tag`? If you didn't want, you'd just call `image_tag`.

meowsus: Brraaaiiiinnnnssss

PR #347, opened 2020-02-01
Add append points for workarea-margin gem

Brandongoodman615:
Allows adding and editing cost to pricing skus.
Shows cost and margin in the admin pricing view.

https://github.com/Brandongoodman615/workarea-margin

### DIFF
--- a/admin/app/views/workarea/admin/pricing_skus/_cards.html.haml
+++ b/admin/app/views/workarea/admin/pricing_skus/_cards.html.haml
@@ -14,6 +14,7 @@
             %li
               %strong #{t('workarea.admin.fields.tax_code')}:
               = model.tax_code.presence || '(none)'
+            = append_partials('admin.card_cost_and_margin', model: model) 
             %li
               %strong #{t('workarea.admin.fields.msrp')}:
               - if model.msrp.blank?
--- a/admin/app/views/workarea/admin/pricing_skus/_cards.html.haml
+++ b/admin/app/views/workarea/admin/pricing_skus/_cards.html.haml
@@ -14,13 +14,6 @@
             %li
               %strong #{t('workarea.admin.fields.tax_code')}:
               = model.tax_code.presence || '(none)'
-            = append_partials('admin.card_cost_and_margin', model: model) 
-            %li
-              %strong #{t('workarea.admin.fields.msrp')}:
-              - if model.msrp.blank?
-                = t('workarea.admin.cards.attributes.no_value')
-              - else
-                = number_to_currency model.msrp
             %li
               %strong #{t('workarea.admin.fields.on_sale')}:
               = model.on_sale?.to_s.titleize
@@ -28,12 +21,18 @@
               %strong #{t('workarea.admin.fields.discountable')}:
               = model.discountable?.to_s.titleize
             %li
+              %strong #{t('workarea.admin.fields.msrp')}:
+              - if model.msrp.blank?
+                = t('workarea.admin.cards.attributes.no_value')
+              - else
+                = number_to_currency model.msrp
+            = append_partials('admin.pricing_sku_attributes_card', pricing_sku: model)
+            %li
               %strong= t('workarea.admin.fields.updated_at')
               #{local_time_ago(model.updated_at)}
             %li
               %strong= t('workarea.admin.fields.created_at')
               #{local_time_ago(model.created_at)}
-            = append_partials('admin.pricing_sku_attributes_card', pricing_sku: model)
 
           = link_to edit_pricing_sku_path(model), class: 'card__button' do
             %span.button.button--small= t('workarea.admin.cards.attributes.button')
--- a/admin/app/views/workarea/admin/pricing_skus/edit.html.haml
+++ b/admin/app/views/workarea/admin/pricing_skus/edit.html.haml
@@ -31,13 +31,6 @@
           = label_tag 'sku[tax_code]', t('workarea.admin.fields.tax_code'), class: 'property__name'
           = text_field_tag 'sku[tax_code]', @sku.tax_code, class: 'text-box'
 
-        = append_partials('admin.add_cost', sku: @sku)
-        
-        .property
-          = label_tag 'sku_msrp', t('workarea.admin.fields.msrp'), class: 'property__name'
-          = currency_symbol
-          = text_field_tag 'sku[msrp]', @sku.msrp, class: 'text-box text-box--small'
-
         .grid
           .grid__cell.grid__cell--50.grid__cell--25-at-medium
             .property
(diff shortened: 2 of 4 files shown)


admin/app/views/workarea/admin/pricing_skus/_cards.html.haml
```diff
@@ -14,6 +14,7 @@
             %li
               %strong #{t('workarea.admin.fields.tax_code')}:
               = model.tax_code.presence || '(none)'
+            = append_partials('admin.card_cost_and_margin', model: model)
```
bencrouse: Looks like we already have an append point for this list at `admin.pricing_sku_attributes_card`. I think in base, we'd consider that sufficient coverage here.

admin/app/views/workarea/admin/pricing_skus/edit.html.haml
```diff
@@ -31,6 +31,8 @@
           = label_tag 'sku[tax_code]', t('workarea.admin.fields.tax_code'), class: 'property__name'
           = text_field_tag 'sku[tax_code]', @sku.tax_code, class: 'text-box'
 
+        = append_partials('admin.add_cost', sku: @sku)
```
bencrouse: Also here, looks like we have `admin.pricing_sku_fields` already in place.

admin/app/views/workarea/admin/pricing_skus/index.html.haml
```diff
@@ -46,6 +46,7 @@
                 = check_box_tag 'select_all', nil, false, id: 'select_all', class: 'checkbox__input', data: { bulk_action_select_all: '' }
                 = label_tag 'select_all', t('workarea.admin.bulk_actions.select_all'), class: 'checkbox__label'
             %th= t('workarea.admin.fields.sku')
+            = append_partials('admin.additional_fields_labels')
```
bencrouse: So, the append names are globally scoped - in this case "additional fields labels" could mean anywhere! Can you use a more specific name like `admin.pricing_skus_additional_field_labels`?

admin/app/views/workarea/admin/pricing_skus/new.html.haml
```diff
@@ -34,6 +34,8 @@
           %span.property__name= t('workarea.admin.fields.discountable')
           = toggle_button_for 'sku[discountable]', @sku.discountable?, title_true: t('workarea.admin.pricing_skus.allow_discounting'), title_false: t('workarea.admin.pricing_skus.disallow_discounting')
 
+        = append_partials('admin.add_cost', sku: @sku)
```
bencrouse: Looks like `admin.pricing_sku_fields` is already here too!

Brandongoodman615:
@bencrouse I will take a look at these as soon as I have a chance. If there are already append points, I will utilize those then.  Also I will update the more generic naming. Thanks for the feedback!

admin/app/views/workarea/admin/pricing_skus/index.html.haml
Brandongoodman615: Agree on this one completely. Will work on implementing the change here.

admin/app/views/workarea/admin/pricing_skus/new.html.haml
Brandongoodman615: I do agree with using the existing append point for the new form as it will look fine with the cost underneath the MSRP.  This also depends on the edit form as I reference in the edit form above.

admin/app/views/workarea/admin/pricing_skus/edit.html.haml
Brandongoodman615: The Edit form throws me off, I think using the same append point that is used in the New form would be fine as long as the MSRP field is moved below On Sale and Allow Discounting, the same way it is laid out in the New form. Picture below for reference, the proposed changes are both using the existing append points.

Screen Shot 2020-02-06 at 7 38 51 PM

admin/app/views/workarea/admin/pricing_skus/_cards.html.haml
Brandongoodman615: The card has a similar layout issue that I have with the edit and new form (posted below). They don't all exactly match in order. The append point is also out of order with the other layouts. This picture it utilizing the current append point in base, which puts the new fields at the bottom and that doesn't make sense to me for those to be in different locations on all three places.

Screen Shot 2020-02-06 at 8 08 52 PM

Brandongoodman615:
To add an over all note for the edit, new, and card views. I think that aligning all the info in the same order (when applicable) would make the most sense. If it were up to me, I would go in the following order across all three...
- Active (this would be sku on new view, I cut it out of the screen shot)
- Tax code
- On Sale / Allow Discounting (side by side for edit and new)
- MSRP
- Append Point
- Updated / Created (side by side in edit)

Would like to add that I'm unaware of how this may affect other gems, or anything else I may be missing or just plain not understanding. Thanks!

Brandongoodman615:
I think we're not on the same page. I can adapt to any append naming convention. my issue is with the append point locations in the codebase are not the same in those three views, like the images I posted along with each comment I made.

Brandongoodman615:
If you would prefer not to move any of the append points, then I'm fine with closing this PR since I have documentation on the gem to add the append points in the proper location where they make the most sense to me.

bencrouse:
I'm down to move append points so long as the move doesn't break HTML. Also, I like the idea of them being consistent. Moving the existing `admin.pricing_sku_attributes_card` is fine with me, it's unlikely anyone would want their new attributes listed below created/updated at.

I'm also open to new append points, so long as they're named generically (not feature specific). I won't merge the branch as-is due to the feature-specific append point naming.

Does this clarify? I hope we can get this sorted out for your plugin!

Brandongoodman615:
I'm still planning to fix naming and utilize the existing append points as long as I can move them in place where they make the most sense. Sounds like you agree on this idea as well. I haven't made any changes from the first just because, honestly, this is the first time I've forked a repo to PR so I get in the weeds a bit making changes, I didn't want to mess with it until we had a good idea of what would be a good path to getting this complete. I will update this with the discussed changes and we can go from there. Thanks!

bencrouse:
Ok sounds great @Brandongoodman615. Thanks for your patience.

Brandongoodman615:
- Uses existing append points in `new`, `edit`, and `card` views
- Specifies append naming for `index` view
- Adds side by side alignment for `on_sale?` and `discountable?` in `new` view to match the `edit` view

**Aligns order for new, edit, and card views** 
1. Active
2. Tax code
3. On Sale / Allow Discounting (side by side for edit and new)
4. MSRP
5. Append Point
6. Updated / Created (side by side in edit)

bencrouse:
Awesome, looks great @Brandongoodman615 .

One last question before I merge - I just noticed this is pointing at `master`, which means it would be released in v3.6.0 (which will probably come out a couple months from now). Did you want to get this in patch release for v3.4 or v3.5 (not sure which version you're working on)? 

If so, wanna update to point at either `v3.4-stable` or `v3.5-stable`? Lemme know, I'd like to get this merged for you!

Brandongoodman615:
I'm really not even sure on this one. I can take a look tonight, hopefully, and try to figure out how to point it to another branch. If you have a preference which one, I can do whatever works best for you. Still trying to understand the whole PRing a forked repo process.

bencrouse:
No problem.

I'd probably just target the version you're working with in your project. Won't introduce unnecessary change for v3.4 if you're working with v3.5, but also not a big deal.

As far as changing the branch, easiest thing would be to start a new branch from whichever stable branch you choose, and then cherry pick this commit onto that new branch. You can then force push that branch to this branch name and the PR will update.

If you just edit the branch on this PR, all the other commits on `master` will be pulled in, which we obviously don't want to merge for the patch branches.

Brandongoodman615:
https://github.com/workarea-commerce/workarea/pull/388
^ new PR for correct branch. 

This should be deleted, leaving it for reference until other branch is good to go.

bencrouse:
Closing, as #388 is merged now.

PR #348, opened 2020-02-03
Add flag to make users tax exempt

jyucis:
Adds a flag to the user model that makes all of that users orders exempt from paying tax.

WORKAREA-198

### DIFF
--- a/admin/app/views/workarea/admin/users/edit.html.haml
+++ b/admin/app/views/workarea/admin/users/edit.html.haml
@@ -68,6 +68,11 @@
               %span.property__name
                 = t('workarea.admin.users.edit.email_signup')
               = toggle_button_for 'email_signup', @user.email_signup?
+          .grid__cell.grid__cell--50.grid__cell--25-at-medium
+            .property
+              = label_tag 'user[tax_exempt]', t('workarea.admin.fields.tax_exempt'), class: 'property__name'
+              = toggle_button_for 'user[tax_exempt]', @user.tax_exempt?, title_true: t('workarea.admin.users.edit.tax_exempt_true'), title_false: t('workarea.admin.users.edit.tax_exempt_false')
+
 
         = append_partials('admin.user_properties', user: @user)
 
--- a/core/app/models/workarea/pricing/calculators/tax_calculator.rb
+++ b/core/app/models/workarea/pricing/calculators/tax_calculator.rb
@@ -3,8 +3,10 @@ module Workarea
     module Calculators
       class TaxCalculator
         include Calculator
+        delegate :tax_exempt?, to: :user, allow_nil: true
 
         def adjust
+          adjust_tax_exempt_items_tax && return if tax_exempt?
           adjust_shipped_items_tax
           adjust_not_shipped_items_tax
         end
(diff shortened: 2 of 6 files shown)


admin/app/views/workarea/admin/users/edit.html.haml
```diff
@@ -36,6 +36,10 @@
           = label_tag 'user[password]', t('workarea.admin.fields.password'), class: 'property__name'
           = password_field_tag 'user[password]', nil, class: 'text-box', autocomplete: 'off'
           %span.property__note= link_to t('workarea.admin.users.edit.send_password_reset'), send_password_reset_user_path(@user), data: { method: :post }
+        .property
```
bencrouse: This is a less-like-to-be-used field, can you add it to the grid with the email signup toggle?

core/app/models/workarea/pricing/calculators/tax_calculator.rb
```diff
@@ -3,8 +3,10 @@ module Pricing
     module Calculators
       class TaxCalculator
         include Calculator
+        delegate :tax_exempt?, to: :user, allow_nil: true
 
         def adjust
+          adjust_tax_exempt_items_tax && return if tax_exempt?
```
bencrouse: I'm wondering about plugin effort, like will the tax 3rd party plugins respect this setting if the logic is in this class?

core/app/models/workarea/pricing/calculators/tax_calculator.rb
```diff
@@ -50,6 +72,15 @@ def not_shipped_items_price_adjustments
...
           shipped_items_price_adjustments
         end
+
+        private
+
+        def user
+          return @user if defined? @user
+          return unless order.user_id.present?
+          @user = User.find(order.user_id)
+        rescue Mongoid::Errors::DocumentNotFound, Mongoid::Errors::InvalidFind
```
bencrouse: This rescue won't cause `@user` to get memoized, so it'll run through the lookup again.

core/app/models/workarea/pricing/calculators/tax_calculator.rb
jyucis: Usually the plugin is responsible for handling tax exemption. 
For example: https://github.com/workarea-commerce/workarea-avatax/blob/master/app/models/workarea/user.decorator

bencrouse: I'm asking because when this originally came up the question was around whether tax plugins supported this. So are we saying we don't want those plugins to respect this setting and go off their own logic?

jyucis: The tax plugins use a an api call to their respective services to return a list of taxable statuses. This list is derived from the  configuration of the 3rd party.

I think we would need to first decorate the B2B plugin to delegate tax_exempt? to the user model. After that we would need decorate the tax plugins user models tax_exempt? method to respect whatever logic they use to indicate if a user is non taxable or not.

bencrouse:
Closing as discussed in the UX meeting.

PR #349, opened 2020-02-03
Fix Hardcoded Currency In System Tests

tubbo:
Refactor system tests to not look for a currency when possible, and when
not possible, read the default currency symbol from `Money` instead of
assuming it is `$`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,5 @@ gem 'workarea-testing', path: 'testing'
 gem 'listen'
 gem 'bundler-audit'
 gem 'rubocop'
-gem 'rails-decorators', git: 'https://github.com/workarea-commerce/rails-decorators.git'
-
-gem 'teaspoon', git: 'https://github.com/jejacks0n/teaspoon.git'
+gem 'rails-decorators', git: 'https://github.com/workarea-commerce/rails-decorators.git', branch: 'master'
+gem 'teaspoon', git: 'https://github.com/jtapia/teaspoon.git', branch: 'chore/update-rails-6'
--- a/admin/test/system/workarea/admin/publish_authorization_system_test.rb
+++ b/admin/test/system/workarea/admin/publish_authorization_system_test.rb
@@ -51,7 +51,7 @@ module Workarea
         click_link 'HTML'
         click_button 'create_block'
 
-        assert(page.has_content?('Success'))
+        assert_text('Success')
         assert(page.has_selector?('.content-block'))
 
         click_link t('workarea.admin.create_catalog_categories.content.continue_to_taxonomy')
--- a/admin/test/system/workarea/admin/publish_authorization_system_test.rb
+++ b/admin/test/system/workarea/admin/publish_authorization_system_test.rb
@@ -51,7 +51,7 @@ module Workarea
         click_link 'HTML'
         click_button 'create_block'
 
-        assert_text('Success')
+        assert(page.has_content?('Success'))
         assert(page.has_selector?('.content-block'))
 
         click_link t('workarea.admin.create_catalog_categories.content.continue_to_taxonomy')
--- a/storefront/test/system/workarea/storefront/categories_system_test.rb
+++ b/storefront/test/system/workarea/storefront/categories_system_test.rb
@@ -3,6 +3,7 @@ require 'test_helper'
 module Workarea
   module Storefront
     class CategoriesSystemTest < Workarea::SystemTest
+      include Storefront::SystemTest
       include BreakpointHelpers
       setup :set_products
       setup :set_search_settings
@@ -62,8 +63,8 @@ module Workarea
 
         assert(page.has_content?('Integration Product 1'))
         assert(page.has_content?('Integration Product 2'))
-        assert(page.has_content?('$10.00'))
-        assert(page.has_content?('$5.00'))
+        assert(page.has_content?('10.00'))
+        assert(page.has_content?('5.00'))
         assert(page.has_content?('Medium (3)'))
         assert(page.has_content?('Small (2)'))
       end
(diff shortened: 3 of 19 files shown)


Gemfile
```diff
@@ -9,5 +9,5 @@ gem 'workarea-testing', path: 'testing'
 gem 'listen'
 gem 'bundler-audit'
 gem 'rubocop'
-gem 'rails-decorators', git: 'https://github.com/workarea-commerce/rails-decorators.git'
+gem 'rails-decorators', git: 'https://github.com/workarea-commerce/rails-decorators.git', branch: 'master'
```
tubbo: I can remove this if you folks want me to, but this makes it a bit easier to work with master branch of workarea since I have a local override on my system for rails-decorators so it's a bit less cumbersome to work with experimental branches across multiple repos.

testing/lib/workarea/storefront/system_test.rb
```diff
@@ -151,6 +151,10 @@ def find_analytics_events(for_event: nil)
           all_events.select { |e| e['name'] == for_event }
         end
       end
+
+      def currency
+        Money.default_currency.symbol
```
tubbo: Wouldn't say I'm "in love" with this approach because I'm not sure how it will jive with multi-site apps that use different currencies per site...

tubbo:
there's one remaining admin test failure: https://github.com/workarea-commerce/workarea/pull/349/checks?check_run_id=424164062

```
Failure:
Workarea::Admin::PublishAuthorizationSystemTest#test_user_cannot_select_publish_now_in_workflows [/Users/tscott/Code/workarea/admin/test/system/workarea/admin/publish_authorization_system_test.rb:54]:
expected to find text "Success" in "Shortcuts\n4 Alerts\nHelp\nwww.example.com\nEnglish\nTest Category: Add Content\nor skip this\nView on Storefront\nTest Category\nAbove Results\nBelow Results\nAdvanced\n← Cancel\nHTML\nSettings\nDisplay\nHTML\n<h3 style=\"font-size: 40px;\">Raw HTML Content</h3>\nPublishing now\nPublish with Publishing Test\na new release\n→\nSave\nCancel\n1) Setup 2) Products 3) Content 4) Taxonomy 5) Publish\nContinue to Taxonomy →"
```

(changed it to `assert_text` so it's actually possible to view)

The screenshot shows there's definitely no flash message. Is this something only I am getting or is anyone else getting this error? I figured master branch might be a bit unstable these days due to the rails 6 thing...

storefront/test/system/workarea/storefront/categories_system_test.rb
```diff
@@ -175,7 +177,8 @@ def test_filtering_products_and_sorting
         visit storefront.category_path(category)
 
         Capybara.match = :first
-        click_link '$10.00 - $19.99 (1)'
+        price_range = "#{currency}10.00 - #{currency}19.99"
```
mttdffy: not all currencies have a prefix symbol, no?

mttdffy: maybe I'm wrong, but in any case, you're doing this in a lot of places, I'd recommend creating a factory method for this.

Gemfile
bencrouse: I don't quite follow - aren't these two lines equivalent given `rails-decorators` default branch is `master`?

storefront/test/system/workarea/storefront/categories_system_test.rb
mttdffy: might be best to find a way to use `number_to_currency` so that behavior of generating this string matches what happens on the storefront

admin/test/system/workarea/admin/publish_authorization_system_test.rb
```diff
@@ -51,7 +51,7 @@ def test_user_cannot_select_publish_now_in_workflows
         click_link 'HTML'
         click_button 'create_block'
 
-        assert(page.has_content?('Success'))
+        assert_text('Success')
```
bencrouse: This is a bit of an unrelated change.

testing/lib/workarea/storefront/system_test.rb
bencrouse: Yeah... I'm not loving that either. Did you try or consider refactoring to make the actual symbol irrelevant? Like just test for the amounts. That seems like it would be cleaner in the end. Most if not all these assertions could changed to do that.

More complex ones could be regex like

```
assert(page.has_content?("Next Day - #{currency}20.00"))
# converts to
assert_match(page.text, /Next Day - .20\.00/)
```

or something like that.

Gemfile
tubbo: @bencrouse sorry, never actually mentioned that I was receiving an error when you run `bundle` if you have a local override set up for the gem and it's pointing to a Git repo. Basically, since the Bundle is depending on `master` branch, the local override needs to also have `master` branch checked out in order to make sure the code is the same.

storefront/test/system/workarea/storefront/categories_system_test.rb
tubbo: I didn't think `number_to_currency` was available in tests...

admin/test/system/workarea/admin/publish_authorization_system_test.rb
tubbo: this was for debugging, the test here failed and i'm still not 100% on why.

testing/lib/workarea/storefront/system_test.rb
tubbo: TIL, `assert_text` takes a regex! :) https://rubydoc.info/github/jnicklas/capybara/master/Capybara/Node/Matchers#has_text%3F-instance_method

tubbo:
I can't figure out how to `click_link` on some of the price filters. I tried the following:

```ruby
click_link '.result-filters__section--price .result-filters__group result-filters__link'
```

and

```ruby
within '.result-filters__section--price .result-filters__group' do
  click_link '.result-filters__link'
end
```

While the `within` call confirmed that the containing elements were being found, selecting the actual link doesn't seem to work.

PR #351, opened 2020-02-06
Use correct page title for navigation menus index

luismerk:
Noticed this while setting up the primary navigation.


bencrouse:
Thanks @luismerk

PR #354, opened 2020-02-06
Drop Support for Ruby 2.4 and Below

tubbo:
If one installs Workarea when running Ruby 2.4.x and below, a test will
fail because it is expecting that `#casecmp?` will be able to handle
`nil` arguments without throwing a `TypeError`. To address this,
`Workarea::Core` now restricts the required Ruby version in the gem
specification to 2.5.0 and above, dropping support for 2.4 and below,
which at this point is also reaching EOL by the Ruby maintainers.

(**NOTE:** For those experiencing deja vu, this was merged and later
reverted in the patch branches, but is being re-introduced in a new
minor version)

WORKAREA-201


bencrouse:
@tubbo actually, it'd be good to mention this in the upgrade guide, can you open a PR to add that?

PR #355, opened 2020-02-06
Fix `#casecmp?` in Traffic Referrer Segment Rule

tubbo:
This method call fails with a `TypeError` on any Ruby version lower than
2.5.0, when the method was changed to support `nil` values in its
arguments. To address this, call `#to_s` on all arguments passed into
the `#casecmp?` calls in `Segment::Rules::TrafficReferrer`. This
prevents an out-of-box test failure if you are running tests on Ruby
2.4.x. Additionally, this commit adds a test matrix to the build that
will ensure all code is tested against the latest versions of Ruby
2.4.x and Ruby 2.6.x, and any other Ruby versions that the platform may
support.

### DIFF
--- a/core/app/models/workarea/segment/rules/traffic_referrer.rb
+++ b/core/app/models/workarea/segment/rules/traffic_referrer.rb
@@ -13,11 +13,11 @@ module Workarea
         end
 
         def medium_match?(referrer)
-          medium.to_s.present? && medium.to_s.strip.casecmp?(referrer.medium.to_s)
+          medium.present? && medium.to_s.strip.casecmp?(referrer.medium.to_s)
         end
 
         def source_match?(referrer)
-          source.any? { |s| s.strip.present? && s.strip.casecmp?(referrer.source.to_s) }
+          source.any? { |s| s.present? && s.to_s.strip.casecmp?(referrer.source.to_s) }
         end
 
         def url_match?(referrer)
--- a/core/app/models/workarea/segment/rules/traffic_referrer.rb
+++ b/core/app/models/workarea/segment/rules/traffic_referrer.rb
@@ -13,11 +13,11 @@ module Workarea
         end
 
         def medium_match?(referrer)
-          medium.to_s.strip.casecmp?(referrer.medium)
+          medium.to_s.present? && medium.to_s.strip.casecmp?(referrer.medium.to_s)
         end
 
         def source_match?(referrer)
-          source.any? { |s| s.strip.casecmp?(referrer.source) }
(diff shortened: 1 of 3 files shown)


core/app/models/workarea/segment/rules/traffic_referrer.rb
```diff
@@ -13,11 +13,11 @@ def qualifies?(visit)
         end
 
         def medium_match?(referrer)
-          medium.to_s.strip.casecmp?(referrer.medium)
+          medium.to_s.present? && medium.to_s.strip.casecmp?(referrer.medium.to_s)
```
bencrouse:
```suggestion
          medium.to_s.strip.casecmp?(referrer.medium.to_s)
```

bencrouse: That extra `to_s` and presence check doesn't get us anything.

core/app/models/workarea/segment/rules/traffic_referrer.rb
```diff
@@ -13,11 +13,11 @@ def qualifies?(visit)
...
         end
 
         def source_match?(referrer)
-          source.any? { |s| s.strip.casecmp?(referrer.source) }
+          source.any? { |s| s.strip.present? && s.strip.casecmp?(referrer.source.to_s) }
```
bencrouse:
```suggestion
          source.any? { |s| s.strip.casecmp?(referrer.source) }
```

bencrouse: That extra `strip` and presence check doesn't get us anything.

core/app/models/workarea/segment/rules/traffic_referrer.rb
tubbo: as discussed in standup, this preserves the logic because `.casecmp` changes behavior in 2.5 to return `false` when both sides are blank, whereas if both sides are blank in 2.4 and below, it will return `true`.

bencrouse:
```suggestion
          medium.present? && medium.to_s.strip.casecmp?(referrer.medium.to_s)
```

core/app/models/workarea/segment/rules/traffic_referrer.rb
bencrouse:
```suggestion
          source.any? { |s| s.present? && s.to_s.strip.casecmp?(referrer.source.to_s) }
```

PR #356, opened 2020-02-07
Provide standardized release task

meowsus:
Behold my legacy!

WORKAREA-163

### DIFF
--- a/core/lib/workarea/release.rake
+++ b/core/lib/workarea/release.rake
@@ -1,0 +1,105 @@
+require 'active_support/inflector'
+
+load 'workarea/changelog.rake'
+
+namespace :workarea do
+  desc "Release the next version of this gem"
+  task :release, [:private] do |task, args|
+    #
+    # Devise current gem version
+    #
+    #
+    version_file_path = Dir['**/version.rb'].reject { |p| p =~ /vendor/ }.first
+    gem_namespace =
+      version_file_path
+        .split('lib/').last
+        .split('/version').first
+        .classify
+
+    gem_name = gem_namespace.gsub(/::/, ' ')
+
+    current_version = if gem_namespace == 'Workarea'
+      "#{gem_namespace}::VERSION::STRING".constantize
+    else
+      "#{gem_namespace}::VERSION".constantize
+    end
+
+    #
+    # Devise next gem version
+    #
+    #
+    current_branch = `git rev-parse --abbrev-ref HEAD`.chomp
+    puts <<~MESSAGE
+      You're about to release a new version of the #{gem_name} gem.
+
+      The current version in the #{current_branch} branch is #{current_version}.
+
+      Please provide the version that you'd like to release:
+    MESSAGE
+
+    next_version = STDIN.gets.chomp
+
+    puts "Should I release #{next_version} from #{current_branch}? (Y/n)"
+    answer = STDIN.gets.chomp
+    answer = 'Y' if answer.empty?
+    abort("Nothing to do") unless answer.casecmp?('Y')
+
+    #
+    # Update version.rb and CHANGELOG.md
+    #
+    #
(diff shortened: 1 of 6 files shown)


core/lib/workarea/release.rake
```diff
@@ -0,0 +1,105 @@
...
+    version_file_path = Dir['**/version.rb'].reject { |p| p =~ /vendor/ }.first
+    gem_namespace =
+      version_file_path
+        .split('lib/').last
+        .split('/version').first
+        .classify
+
+    gem_name = gem_namespace.gsub(/::/, ' ')
+
+    current_version = if gem_namespace == 'Workarea'
```
bencrouse: I'm kinda confused on all this stuff to get the version - I believe in the all gems if you just load the `whatever.gemspec`, it will `require` the right `lib/*/version.rb` file and then you can reference that. Am I missing something?

core/lib/workarea/release.rake
```diff
@@ -0,0 +1,105 @@
...
+      git commit -m 'Release version #{next_version}' &&
+      git tag -a 'v#{next_version}' -m 'Tagging #{next_version}' &&
+      git push origin HEAD --follow-tags
+    COMMAND
+
+    #
+    # Build gems, push to hosts, and clean up
+    #
+    #
+    hosts = ["https://#{ENV['BUNDLE_GEMS__WORKAREA__COM']}@gems.workarea.com"]
```
bencrouse: This will prob output those, which feels unnecessary and potentially insecure?

core/lib/workarea/release.rake
```diff
@@ -0,0 +1,105 @@
...
+
+    #
+    # Update version.rb and CHANGELOG.md
+    #
+    #
+    tempfile = Tempfile.new('version_temp.rb')
+    File.open(version_file_path, 'r') do |file|
+      file.each_line do |line|
+        tempfile.puts line.include?(current_version) \
+          ? line.gsub(current_version, next_version) \
```
bencrouse: Is this reliable? Feels... I dunno. Unreliable or something.

core/lib/workarea/version.rb
```diff
@@ -1,11 +1,6 @@
 module Workarea
   module VERSION
-    MAJOR = 3
```
bencrouse: I'm not sure why but just ditching these constants seems dangerous.

core/lib/workarea/release.rake
meowsus: This is how we were always doing it. I could investigate other options, but this is pretty much a lateral move from what we had before.

core/lib/workarea/release.rake
meowsus: This is pretty reliable. Now with the change to base's `version.rb` files all gems have a consistent string that will be replaced during the update.

core/lib/workarea/version.rb
meowsus: We don't use the constants individually except in the github action to build documentation. I just updated that file to reference the new format. The goal here is to bolster the reliability of the gsubbing the version in the rake task.

core/lib/workarea/release.rake
meowsus: I think I went this route because gems are likely to have one relevant `version.rb` file, but could have many `gemspec`s. Plus I needed the version file anyway, so grabbing the version out of there seemed reasonable.

core/lib/workarea/release.rake
tubbo: instead of parsing the `version.rb` file, we should just be reading the `Workarea::GemName::VERSION` constant that should already be loaded in Rakefile.

bencrouse:
I'm closing this for now, to clean up our PR list. We'll reopen after we circle back to the release process.

PR #357, opened 2020-02-11
Add Redis cache store change to v3.5 upgrade guide

bencrouse:
Several builds have hit this in their upgrade.

No changelog

WORKAREA-170


tubbo:
Totally cool with this, of course. Were people having problems using redis-activesupport with 3.5 or is this just a “consistency” thing?

bencrouse:
@tubbo we took out the gem dependency. The Rails oob store has a different name and way of taking options, so if you upgrade, you have to change it. Courtesy of @MMartyn .

PR #358, opened 2020-02-11
Fix Overwriting CORS Rules for S3 Direct Uploads

tubbo:
Workarea previously replaced the existing CORS configuration on the S3
bucket used for storing direct uploads with its own, which caused issues
for environments that share an S3 bucket between servers (such as ad-hoc
demo servers or low-traffic "all-in-one" instances). Instead of
replacing the entire configuration, Workarea now reads the existing
allowed hosts configuration and appends its own onto the end, preserving
the configuration that previously existed. This should address the
problem wherein if another server attempts a direct upload, it can
revoke the access from previous servers to upload to the S3 bucket,
since they were no longer in the CORS configuration.


tubbo:
whoops, forgot to mock it out in the system test. should be OK now.

PR #359, opened 2020-02-12
Handle Deleted Categories in Category Options

tubbo:
In the `options_for_category` method, Workarea did not previously check
for whether a category exists, resulting in Mongoid throwing a
`DocumentNotFound` error when encountering the method and causing a 500
error in the real world. This has now been resolved by rewriting the
code to check for whether the model was found before proceeding.
`options_for_category` will now return `nil` early when this occurs.

### DIFF
--- a/admin/app/helpers/workarea/admin/content_helper.rb
+++ b/admin/app/helpers/workarea/admin/content_helper.rb
@@ -36,7 +36,7 @@ module Workarea
       def options_for_category(category_id)
         model = Catalog::Category.where(id: category_id).first
 
-        return unless model.present?
+        return '' unless model.present?
 
         options_for_select({ model.name => model.id }, model.id)
       end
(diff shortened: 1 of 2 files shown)


admin/app/helpers/workarea/admin/content_helper.rb
```diff
@@ -34,9 +34,10 @@ def storefront_content_preview_path(content)
       end
 
       def options_for_category(category_id)
-        return nil unless category_id.present?
+        model = Catalog::Category.where(id: category_id).first
+
+        return '' unless model.present?
```
bencrouse: We can still return a valid HTML list of options here, if you do something like `return options_for_select({}) unless model.present?`

tubbo: that actually returns the same thing :)

```
[2] pry(main)> include ActionView::Helpers
=> Object
[3] pry(main)> options_for_select({})
=> ""

PR #360, opened 2020-02-13
Don't store content block types in configuration

bencrouse:
Storing these in `Workarea.config` breaks when combined with the
multisite plugin because `Workarea.config` gets copied to the site as
part of the creating the site. Since creating the site happens during
initialization and the blocks typesaren't set on `Workarea.config`, you
end up with an empty configuration for block types.

Storing these elsewhere ensures they are set regardless of how
`Workarea.config` has been copied during initialization.

Reported in Disource here: https://discourse.workarea.com/t/multi-site-and-testing/1778/3

### DIFF
--- a/admin/app/controllers/workarea/admin/content_blocks_controller.rb
+++ b/admin/app/controllers/workarea/admin/content_blocks_controller.rb
@@ -9,7 +9,7 @@ module Workarea
     def index
       @return = params[:return_to].to_h
 
-      @system_block_types = Workarea.config.content_block_types
+      @system_block_types = Configuration::ContentBlocks.types
       @content_presets = Content::Preset.all.to_a
     end
 
--- a/core/app/models/workarea/content/block.rb
+++ b/core/app/models/workarea/content/block.rb
@@ -96,7 +96,7 @@ module Workarea
       # @return [Workarea::Content::BlockType]
       #
       def type
-        Workarea.config.content_block_types.detect { |bt| bt.id == type_id }
+        Configuration::ContentBlocks.types.detect { |bt| bt.id == type_id }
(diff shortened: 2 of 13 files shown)


tubbo:
Removing this config in a patch seems dangerous. I see some usage of it in private repos, and [we even depend on it in OneTheme](https://github.com/workarea-commerce/workarea-one-theme/blob/master/app/seeds/workarea/content_block_page_seeds.rb). Is there any way to notify the developer that this configuration has been deprecated, and they need to change where they put it?

bencrouse:
That's a good point, a quick search shows several projects referencing it too. We could copy the value in there after initialization to help, maybe? But if it's being referenced during initialization, the problem would still exist.

Any other ideas on how to fix?

bencrouse:
I added a line to copy the values to `Workarea.config` after they're loaded. But as far as deprecation, that would be nice, any thoughts on a good way to do that?

tubbo:
```ruby
# The content block types currently available. Most easily customized by
# calling Workarea.define_content_block_types
config.deprecate! :content_block_types, SwappableList.new, 'Use Configuration::ContentBlocks.types instead'
```

```ruby
module Workarea
  module Configuration
    class AdministrableOptions < ActiveSupport::InheritableOptions
      def deprecate!(name, value, message)
        Workarea.deprecation.deprecate_methods(self, name => message)
        self[name] = value
      end

      # ...
    end
  end
end
```

bencrouse:
Ok I'll give that a shot when I get back.

bencrouse:
Alright, I added the deprecation warning. Going to merge so I can get some builds going.

PR #362, opened 2020-02-19
Fix Regular Price Displays In SKUs

tubbo:
The **Regular Price** value in admin was being aliased to the
`#sell_price`, which can sometimes be the sale price. To remedy this,
Workarea now explicitly defines the `Pricing::Sku#regular_price` method
and defines `Admin::PricingSkuViewModel#on_sale?` to represent whether
any of the prices in a SKU are on sale. The SKU is considered "On Sale"
if this is so.

**Before:**

Screen Shot 2020-02-18 at 5 17 49 PM

**After:**

Screen Shot 2020-02-18 at 6 15 50 PM

Sorry I didn't screencap the "On Sale" column, but you can see what that looked like if you take a gander in the ticket.

### DIFF
--- a/admin/app/view_models/workarea/admin/pricing_sku_view_model.rb
+++ b/admin/app/view_models/workarea/admin/pricing_sku_view_model.rb
@@ -16,6 +16,10 @@ module Workarea
       def inventory
         @inventory ||= Inventory::Sku.where(id: model.id).first
       end
+
+      def on_sale?
+        prices.any?(&:on_sale?)
+      end
     end
   end
 end
--- a/core/app/models/workarea/pricing/sku.rb
+++ b/core/app/models/workarea/pricing/sku.rb
@@ -47,14 +47,22 @@ module Workarea
       end
 
       # Default selling price, when one unit is placed
-      # into the cart of an anonymous user.
+      # into the cart of an anonymous user. Determines whether the SKU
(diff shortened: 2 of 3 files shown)


admin/app/view_models/workarea/admin/pricing_sku_view_model.rb
```diff
@@ -16,6 +16,10 @@ def product
       def inventory
         @inventory ||= Inventory::Sku.where(id: model.id).first
       end
+
+      def on_sale?
```
tubbo: This will display "On Sale: Yes" for SKUs in the index where any prices are marked as "On Sale".

core/app/models/workarea/pricing/sku.rb
```diff
@@ -47,14 +47,22 @@ def find_price(options = {})
...
+      # into the cart of an anonymous user. Determines whether the SKU
+      # is on sale and returns the `:sale` price if so, otherwise it
+      # will return the `:sell` price.
       #
       # @return [Money]
       #
       def sell_price
         find_price(quantity: 1).sell
       end
-      alias_method :regular_price, :sell_price
```
bencrouse: Isn't this a breaking change? Feels like this could really break something in a build. If this is a display issue, it would sound like we could fix it without changing the definition of this one method.

PR #363, opened 2020-02-19
Return a link to a github page when plugin match is found in search.

jyucis:
WORKAREA-179

### DIFF
--- a/docs/source/javascripts/site.js
+++ b/docs/source/javascripts/site.js
@@ -198,7 +198,6 @@ window.modules.search = (function () {
             var plugin, list = '';
 
             $.each(results, function (i, plugin) {
-                plugin = results[i];
                 list += pluginResultListItemTPL(plugin.displayName, plugin.gitHubKey);
             });
             return list;
@@ -261,8 +260,8 @@ window.modules.search = (function () {
         bindToSearchInput = function (lunr) {
             $('#lunr-search').on('keyup', function (event) {
                 var results = lunr.index.search(event.target.value);
-                var plugin_result = pluginResult(event.target.value);
-                display(results, plugin_result, lunr);
+                var plugin = pluginResult(event.target.value);
+                display(results, plugin, lunr);
             });
         },
 
--- a/docs/source/javascripts/site.js
+++ b/docs/source/javascripts/site.js
@@ -13,6 +13,7 @@ const plugins = [
     },
     {
         matches: ['afterpay', 'after pay'],
+        displayName: 'Afterpay',
         gitHubKey: 'workarea-afterpay'
     },
     {
--- a/docs/source/javascripts/site.js
+++ b/docs/source/javascripts/site.js
@@ -1,5 +1,116 @@
 'use strict';
 
+const plugins = [
+    {
+        matches: ['moneris'],
+        displayName: 'Moneris',
+        gitHubKey: 'workarea-moneris'
+    },
+    {
+        matches: ['paypal', 'pay pal'],
+        displayName: 'PayPal',
+        gitHubKey: 'workarea-paypal'
+    },
+    {
+        matches: ['afterpay', 'after pay'],
+        gitHubKey: 'workarea-afterpay'
+    },
+    {
+        matches: ['emarsys'],
+        displayName: 'Emarsys',
+        gitHubKey: 'workarea-emarsys'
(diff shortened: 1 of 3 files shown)


docs/source/javascripts/site.js
```diff
@@ -71,6 +82,43 @@ window.modules.search = (function () {
...
+                list += pluginResultListItemTPL(plugin.name, plugin.git_hub_key);
+            }
+
+            return list;
+        },
+
+        pluginResult = function (query) {
+            var results = [],
+                plugins = [
+                    {
```
jyucis: RFC: A better place to put this list?

tubbo: It would be _really_ neat if we could pull JSON results from https://plugins.workarea.com, but I just checked and it doesn't seem there's a JSON response for `/plugins`...

docs/source/javascripts/site.js
```diff
@@ -60,6 +60,17 @@ window.modules.search = (function () {
             return item;
         },
 
+        pluginResultListItemTPL = function (name, git_hub_key) {
+            var item = '';
+            item += '<p class="search-results__title"> Looking for the ' + name + " plugin? - ";
```
tubbo: Shouldn't be linking them to https://plugins.workarea.com rather than GitHub? This would prevent searching for proprietary plugins.

jyucis: I dont think plugins isnt building anymore, so github seemed like the save option.

docs/source/javascripts/site.js
```diff
@@ -71,6 +82,43 @@ window.modules.search = (function () {
             return list;
         },
 
+        constructPluginResultList = function (results) {
+            var plugin, list = '';
+            for (var i = 0; i < results.length; i++) {
```
bencrouse: FWIW, could have used `$.each(results, function (i, plugin) { /* ... */ });`

docs/source/javascripts/site.js
bencrouse: How about using a constant-like var in this file? So at least it won't make this function super long to read.

docs/source/javascripts/site.js
```diff
@@ -71,6 +82,43 @@ window.modules.search = (function () {
...
+                        name: 'paypal',
+                        git_hub_key: 'workarea-paypal'
+                    },
+                    {
+                        name: 'afterpay',
+                        git_hub_key: 'workarea-afterpay'
+                    }
+                ]
+
+            for (var i=0; i < plugins.length; i++) {
```
bencrouse: `$.each`...

docs/source/javascripts/site.js
```diff
@@ -81,11 +129,19 @@ window.modules.search = (function () {
             return message;
         },
 
-        display = function (results, lunr) {
+        display = function (results, plugin_result, lunr) {
```
bencrouse: Prefer `camelCase` in JS.

docs/source/javascripts/site.js
```diff
@@ -71,6 +194,31 @@ window.modules.search = (function () {
             return list;
         },
 
+        constructPluginResultList = function (results) {
+            var plugin, list = '';
+
+            $.each(results, function (i, plugin) {
+                plugin = results[i];
```
bencrouse: Don't need this line, `plugin` is being passed in as a result of using `$.each`

docs/source/javascripts/site.js
```diff
@@ -105,7 +261,8 @@ window.modules.search = (function () {
         bindToSearchInput = function (lunr) {
             $('#lunr-search').on('keyup', function (event) {
                 var results = lunr.index.search(event.target.value);
-                display(results, lunr);
+                var plugin_result = pluginResult(event.target.value);
```
bencrouse: Don't use `snake_case` in JS, prefer `camelCase`. ESLint must not be checking this dir, we should fix that at some point.

docs/source/javascripts/site.js
tubbo: I'm not downvoting your solution, I'm downvoting the fact that plugins no longer builds READMEs...

bencrouse:
@jyucis docs deploys happen against `v3.5-stable` right now, can you update to point at that branch?

PR #367, opened 2020-02-21
Fix S3 CORS in tests

bencrouse:
It's annoying and unnecessary to have to stub this for every test that
uses an asset picker.

WORKAREA-209


pachkovsky:
@bencrouse this PR is mentioned in 3.4.27 release notes but it haven't made it to 3.4 branch, is it planned to?

bencrouse:
Hey @pachkovsky that looks like an oversight, we'll get that into the v3.4 patch release next week. Thanks for reporting.

PR #368, opened 2020-02-21
Remove Changes Count in Releases Index

tubbo:
The `#changesets_for_releasable` query cannot be optimized any further
without using some kind of aggregation. Remove it from the index so it
won't cause performance problems.

### DIFF
--- a/admin/config/locales/en.yml
+++ b/admin/config/locales/en.yml
@@ -1516,7 +1516,6 @@ en:
         category: Category
         category_id: Category
         category_ids: Categories
-        changes: Changes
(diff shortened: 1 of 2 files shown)


admin/config/locales/en.yml
```diff
@@ -1516,7 +1516,6 @@ en:
         category: Category
         category_id: Category
         category_ids: Categories
-        changes: Changes
```
tubbo: yes, I did check whether this was used elsewhere before I deleted it. thanks for asking.

PR #371, opened 2020-02-21
Fixing Dragonfly Configuration Doc Page URL

benparsell:
The URL for the Dragonfly Config Docs had an extra forward slash at the end causing a 404.


bencrouse:
Hey @benparsell it looks like maybe you branched from a different branch than you're opening the PR against?

Thanks for opening! Will merge when it's cleaned up.

benparsell:
Ah, shoot. I must have started that branch off from master. I'm going to close and open a separate PR using v3.5-stable, because I don't believe I can change the source branch base.

PR #382, opened 2020-03-02
Fix BSON Bug

tubbo:
Return the `BSON::Time#to_bson` method back to its original
implementation to solve the failures in metrics, reporting, and insights
tests. This is a temporary measure until we can fully solve the issue with the
MongoDB team.

**More Info:** https://jira.mongodb.org/browse/RUBY-2145

### DIFF
--- a/core/lib/workarea/ext/freedom_patches/bson.rb
+++ b/core/lib/workarea/ext/freedom_patches/bson.rb
@@ -8,4 +8,10 @@ module BSON
       to_s
     end
   end
+
+  module Time
+    def to_bson(buffer = ByteBuffer.new, validating_keys = Config.validating_keys?)
+      buffer.put_int64((to_i * 1000) + (usec / 1000))
+    end
+  end
 end
--- a/core/lib/workarea/ext/freedom_patches/bson.rb
+++ b/core/lib/workarea/ext/freedom_patches/bson.rb
@@ -10,6 +10,8 @@ module BSON
   end
 
   module Time
+    # This should be removed when
+    # https://jira.mongodb.org/browse/RUBY-2145 is released.
     def to_bson(buffer = ByteBuffer.new, validating_keys = Config.validating_keys?)
       buffer.put_int64((to_i * 1000) + (usec / 1000))
     end
--- a/core/lib/workarea/core.rb
+++ b/core/lib/workarea/core.rb
@@ -26,6 +26,7 @@ end
 #
 #
 require 'mongoid'
+require 'bson/active_support'
 require 'sidekiq'
 require 'sidekiq/web'
 require 'sidekiq/cron'


core/lib/workarea/ext/freedom_patches/bson.rb
```diff
@@ -8,4 +8,10 @@ def as_json(*args)
       to_s
     end
   end
+
+  module Time
+    def to_bson(buffer = ByteBuffer.new, validating_keys = Config.validating_keys?)
+      buffer.put_int64((to_i * 1000) + (usec / 1000))
```
mttdffy: we're talking about 1 millisecond difference between the old and new implementation. The new one being more accurate...theoretically. Seems like there might an consistency somewhere in our implementation where 1 millisecond makes our tests break. no?

mttdffy: Seems like there is something else going on. This failure for example:

```
Failure:
Workarea::Reports::SalesBySkuTest#test_time_zones [/home/runner/work/workarea/workarea/core/test/queries/workarea/reports/sales_by_sku_test.rb:138]:
Failed on UTC.
Expected: 1
  Actual: 3
```

The test only generates 2 records. How is it getting 3 results? Are you sure there isn't a change that caused the database to not clear or something?

mttdffy: ok. I think I am on board. 

```ruby
today = Time.current.beginning_of_day.utc
# => 2020-03-02 00:00:00 UTC
tomo = (Time.current + 1.day).^C
# ^C
# end_of_today = Time.current.end_of_day.utc
# => 2020-03-02 23:59:59 UTC
start_of_tomo = (Time.current + 1.day).beginning_of_day.utc
# => 2020-03-03 00:00:00 UTC
end_of_today.to_f
# => 1583193600.0
start_of_tomo.to_f
# => 1583193600.0
(end_of_today.to_i * 1000) + (end_of_today.usec / 1000)
# => 1583193599999
(end_of_today.to_f * 1000).round
# => 1583193600000
(start_of_tomo.to_f * 1000).round
# => 1583193600000
```
essentially that 1 millisecond rounding causes the end of today to be equal to start of tomorrow. And thats our problem.

mttdffy: though, it does look like we should require `bson/active_support`.

mttdffy: though, it does look like we should require `bson/active_support`.

mttdffy: can you also add a comment with a link to the issue they commented about so we can keep track of when it's fixed and can take this out.

tubbo: @mttdffy i'm working on adding `bson/active_support` right now, just got a reply on the ticket.

tubbo: https://jira.mongodb.org/browse/RUBY-2145

tubbo: also, I tried adding `bson/active_support` and removing the freedom patch, but still the same issue. apparently we'll need this regardless though...nice of them to tell us.

mttdffy: yeah, the active_support will ensure it works on TimeWithZone but we operate mostly on Time directly in the reports.

bencrouse:
This was fixed and released here: https://github.com/mongodb/bson-ruby/releases/tag/v4.8.1 so I'm closing this. Thanks for looking into it @tubbo .

PR #383, opened 2020-03-02
Prevent Tracking Index Filters on JSON Requests

tubbo:
When `.json` requests are made against the admin, the
`#track_index_filters` callback was previously saving off the full path,
resulting in issues with the back-linking on the admin UI. To resolve
this, Workarea no longer considers `.json` requests on the index page to
be a valid `session[:last_index_path]`.

### DIFF
--- a/admin/app/controllers/workarea/admin/application_controller.rb
+++ b/admin/app/controllers/workarea/admin/application_controller.rb
@@ -102,7 +102,7 @@ module Workarea
       end
 
       def track_index_filters
-        session[:last_index_path] = request.fullpath
+        session[:last_index_path] = request.fullpath unless request.xhr? || request.format.json?
       end
     end
   end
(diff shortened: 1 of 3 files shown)


admin/app/controllers/workarea/admin/application_controller.rb
```diff
@@ -102,7 +102,7 @@ def setup_alerts
       end
 
       def track_index_filters
-        session[:last_index_path] = request.fullpath
+        session[:last_index_path] = request.fullpath unless request.format.json?
```
mttdffy: i know this is technically more encompassing than `request.xhr?` but do we ever make json requests not from xhr? It would be more consistent with logic in other places.

bencrouse: Agreed, when we talked about this at the sprint planning we said we should check both JSON and XHR and ignore for both.

PR #389, opened 2020-03-05
Fix inline Sidekiq workers not running Sidekiq middleware

bencrouse:
This could cause unexpected results if you're expecting middleware to do
something. Whether a job is inline or not should be invisible to the
middleware/job.


bencrouse:
Spent a while trying to figure out a test, but I'm moving on unless someone feels strongly.

bencrouse:
Actually, this doesn't run the client middleware :rage1: 

Ended up doing things differently so this bug didn't matter, so I may close this. RFC.

bencrouse:
Closing for now, until I have time to come back and implement the client side middleware too.

PR #394, opened 2020-03-16
Handle missing price in sell_price method itself

tubbo:
Fixes QA issue.

### DIFF
--- a/admin/app/view_models/workarea/admin/pricing_sku_view_model.rb
+++ b/admin/app/view_models/workarea/admin/pricing_sku_view_model.rb
@@ -62,7 +62,7 @@ module Workarea
       #
       # @return [String]
       def sell_price
-        return 0.to_m.format if min_price.blank?
+        return if min_price.blank?
         return min_price.format unless show_range?
 
         "#{max_price.format} – #{min_price.format}"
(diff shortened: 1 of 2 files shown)


admin/app/view_models/workarea/admin/pricing_sku_view_model.rb
```diff
@@ -62,7 +62,9 @@ def on_sale?
       #
       # @return [String]
       def sell_price
+        return 0.to_m.format if min_price.blank?
```
bencrouse: Selling it for `$0` is meaningfully different than not having a price specified. Let's return `nil` instead and handle it with a en dash or something in display to indicate it's blank.

admin/app/views/workarea/admin/pricing_skus/index.html.haml
```diff
@@ -64,7 +64,7 @@
                 = upcoming_changesets_icon_for(result)
               = append_partials('admin.pricing_skus_additional_fields', result: result)
               %td.align-right= number_to_currency result.msrp
-              %td.align-right= result.sell_price
+              %td.align-right= result.sell_price || '-'
```
bencrouse:
```suggestion
              %td.align-right= result.sell_price || '–'
```

PR #414, opened 2020-04-15
Fix locale specific indexing

bencrouse:


bencrouse:
Note to self, did this on the wrong branch, cherry-pick into `v3.4-stable` when it's merged.

PR #419, opened 2020-04-21
Fix fulfillment shipped mailer template

heyqule:
This commit fixes that fulfillment shipped mailer template is using cancellation header.


bencrouse:
Hey @heyqule I'm going to cherry-pick this into `v3.4-stable` branch so it makes into the patch releases. In the future, point your branch at the minor version branch you're looking for. Thanks for contributing!

PR #420, opened 2020-04-24
Fix pagination bugs

bencrouse:


storefront/app/view_models/workarea/storefront/category_view_model.rb
```diff
@@ -27,7 +27,7 @@ def browser_title
 
       def search_query
         @search_query ||= Search::CategoryBrowse.new(
-          options.merge(
+          options.except(:per_page).merge(
```
gajendrajena, line 30: this restricts the helper to inject ```:per_page```, any specific reason for not allowing ```:perpage```?

bencrouse: Correct, I'm pasting in my message to you on the support ticket:

Hi Gajendra,
It looks like this is a bug in the query page calculation. We'll fix this in the next patch release, but you should be aware that we don't intend page size to be exposed to the outside world, so that will be part of the fix. Larger page sizes is one of the most harmful things you can do for performance, so we intend that to be locked down (only be controlled within the app config). This is a main reason we implement loading more pages on a single screen via JS since customers want to see as many things on a single page as possible.
I'll work on the fix today and post an update to the PR so you can choose how to proceed for your app.
Thanks, Benjamin

PR #425, opened 2020-05-04
Remove extra order ID cookie

bencrouse:
No need for the extra cookie if the order isn't persisted. Note this
doesn't actually affect functionality.

### DIFF
--- a/storefront/app/controllers/workarea/storefront/current_checkout.rb
+++ b/storefront/app/controllers/workarea/storefront/current_checkout.rb
@@ -26,7 +26,10 @@ module Workarea
       #
       def current_order=(order)
         @current_order = order
-        cookies.permanent.signed[:order_id] = order&.id
+
+        if order.blank? || order.persisted?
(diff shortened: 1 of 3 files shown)


storefront/app/controllers/workarea/storefront/current_checkout.rb
```diff
@@ -26,7 +26,10 @@ def current_order
       #
       def current_order=(order)
         @current_order = order
-        cookies.permanent.signed[:order_id] = order&.id
+
+        if order.blank? || order.persisted?
```
MMartyn, line 30: I am confused by this. Did you mean !order.blank?

MMartyn: Or maybe it is a way to unset it when you need to?

bencrouse: Yeah, it was intentional because if you set `current_order = nil`, we want to clear out that cookie. There is some behavior that relies on that.

PR #429, opened 2020-05-18
Improve segment consistency

bencrouse:


storefront/test/integration/workarea/storefront/checkouts_integration_test.rb
```diff
@@ -345,6 +345,12 @@ def test_no_valid_shipping_options
 
         assert(flash[:error].present?)
       end
+
+      def test_not_setting_a_blank_tracking_email
```
MMartyn, line 349: Is this test right? It sets a cookie and tests that it is still set?

bencrouse: Yes, without the change in the checkouts controller, it will blank out the tracking email incorrectly.

bencrouse: The test is to assert the email doesn't get cleared.

PR #430, opened 2020-05-18
Remove caching from direct upload CORS requests

bencrouse:
The caching continues to give us problems, and this isn't a high-traffic
part of the system so there isn't a practical need for it.


core/app/services/workarea/direct_upload.rb
```diff
@@ -21,7 +18,6 @@ def self.ensure_cors!(request_url)
       cors['CORSConfiguration'].uniq!
 
       Workarea.s3.put_bucket_cors(Configuration::S3.bucket, cors)
```
MMartyn, line 20: Could compare the cors['CORSConfiguration'] to the original and if changed update the bucket to cut down on unneeded calls.

PR #436, opened 2020-05-29
Changes to support package product kits

mttdffy:


admin/app/view_models/workarea/admin/variant_view_model.rb
```diff
@@ -12,7 +12,7 @@ def inventory
     end
 
     def available_inventory
-      inventory.available
+      inventory.available_to_sell
```
bencrouse, line 15: This seems like changing the meaning of this method, what was the idea?

mttdffy: i honestly don't know for sure. 3 months is a long time. Looking at where it gets used I assume its a display thing in the admin, but would cause problems so I'll take it out and see if there is another change that can be made.

bencrouse:
I'm just kind of waiting on this one just until the package products plugin more or less is wrapped up in case additional changes are needed.

PR #438, opened 2020-06-04
Add index for releasable fields on changets, correct order fraud index

mttdffy:
WORKAREA-235

### DIFF
--- a/core/app/models/workarea/order.rb
+++ b/core/app/models/workarea/order.rb
@@ -47,13 +47,13 @@ module Workarea
       {
         placed_at: 1,
         reminded_at: 1,
+        fraud_suspected_at: 1,
         checkout_started_at: 1,
         email: 1,
-        "items[0]._id": 1,
-        fraud_suspected_at: 1
+        "items[0]._id": 1
       },
       {
-        name: 'abandoned_order_email_with_fraud_index',
+        name: 'abandoned_order_email_with_fraud_index_v2',
         background: true
       }
     )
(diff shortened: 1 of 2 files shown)


core/app/models/workarea/order.rb
```diff
@@ -47,10 +47,10 @@ class Order
...
         reminded_at: 1,
+        fraud_suspected_at: 1,
         checkout_started_at: 1,
         email: 1,
-        "items[0]._id": 1,
-        fraud_suspected_at: 1
+        "items[0]._id": 1
       },
       {
         name: 'abandoned_order_email_with_fraud_index',
```
bencrouse, line 56: Ok, so one thing we learned while you were out was that if we change an existing index, and a build upgrades and tries to run the create indexes task, it barfs with an error like this:

```
Mongo::Error::OperationFailure: Index must have unique name.The existing index: { v: 2, key: { placed_at: 1, reminded_at: 1, checkout_started_at: 1, email: 1, items[0]._id: 1, fraud_suspected_at: 1 }, name: "abandoned_order_email_with_fraud_index", ns: "my_app_development.workarea_orders", background: true } has the same name as the requested index: { v: 2, key: { placed_at: 1, reminded_at: 1, fraud_suspected_at: 1, checkout_started_at: 1, email: 1, items[0]._id: 1 }, name: "abandoned_order_email_with_fraud_index", ns: "my_app_development.workarea_orders", background: true } (86) (on localhost:27017)
```

So, can you update the name? Maybe like append a `_v2` or something, not sure how else to handle it but def open to ideas.

PR #448, opened 2020-06-17
Fix Segments Workflow Setup Duplication

tubbo:
The setup form for the new custom segment workflow did not include the
ID of an existing segment (if persisted) in the form when submitted,
causing multiple duplicate segment records to be created when users go
back to the setup step in the workflow. None of the other steps are
affected because the ID appears in the URL, but the setup step does a
direct POST to `/admin/create_segments`, thus causing this problem.

### DIFF
--- a/admin/app/views/workarea/admin/create_segments/setup.html.haml
+++ b/admin/app/views/workarea/admin/create_segments/setup.html.haml
@@ -10,6 +10,8 @@
       = render_message 'error', message
 
     = form_tag create_segments_path, method: 'post' do
+      = hidden_field_tag :id, @segment.id unless @segment.new_record?
+
       .section
         .property.property--required
           = label_tag 'segment_name', t('workarea.admin.fields.name'), class: 'property__name'


admin/test/system/workarea/admin/create_segments_system_test.rb
```diff
@@ -0,0 +1,37 @@
+require 'test_helper'
+
+module Workarea
+  module Admin
+    class CreateSegmentsSystemTest < Workarea::SystemTest
```
tubbo, line 5: This kinda needs to be a system test because we're testing the presence of an `<input type="hidden">` in the form. Anything else would, in my opinion, be mocking out things too much and could result in a regression without failing the test suite.

admin/app/views/workarea/admin/create_segments/setup.html.haml
```diff
@@ -10,6 +10,8 @@
       = render_message 'error', message
 
     = form_tag create_segments_path, method: 'post' do
+      = hidden_field_tag :id, @segment.id unless @segment.new_record?
```
tubbo, line 13: We can also stuff this in the URL, but this seems like the way we're doing it everywhere else, so I figured why rock the boat.

admin/test/system/workarea/admin/create_segments_system_test.rb
bencrouse: I agree, but it also seems heavy handed to add an entire system test for one hidden field. I'm trying to be more sensitive to the performance costs of system tests, so I'd prefer to either take it out or use an `assert_select` in an integration test.

tubbo: I'm just gonna remove it, not sure if constantly having a test for this is worth rethinking everything.

PR #451, opened 2020-06-18
Use Display Name For Applied Facet Values

tubbo:
When rendering the applied filters, wrap the given facet value in
the `facet_value_display_name` helper, ensuring that the value rendered
is always human readable. This addresses an issue where if the applied
filter value is that of a BSON ID, referencing a model somewhere, the
BSON ID was rendered in place of the model's name.


tubbo:
@bencrouse this is the 3.4 version

PR #455, opened 2020-06-23
Raise Error When URL Not Allowed Through Proxy

tubbo:
The proxy in Commerce Cloud will return a 403 response with a special header when a request has not been allowed through. In this case, Workarea now raises an error describing the problem and how to go about solving it. This resolves an issue where previously, errors thrown as a result of a misconfigured proxy would be cryptic and hard to debug. With this more descriptive error, developers will be able to see from Sentry how they can resolve such a scenario.

Here's what someone might see if they trigger this error:

```
Dragonfly::Job::FetchUrl::ProxyError: Error fetching 'http://via.placeholder.com/350x150' with Dragonfly!

This URL is not allowed through the proxy, and therefore the
request was blocked by the Workarea Commerce Cloud firewall.

To allow the host through your proxy server, run the following
command on your local development workstation:

    workarea [qa|staging|production] edit proxy

Thanks for using Commerce Cloud, and keeping your Workarea
application safe.
```

### DIFF
--- a/core/lib/workarea/ext/freedom_patches/dragonfly_job_fetch_url.rb
+++ b/core/lib/workarea/ext/freedom_patches/dragonfly_job_fetch_url.rb
@@ -1,6 +1,25 @@
 module Dragonfly
   class Job
     class FetchUrl < Step
+      class ProxyError < RuntimeError
+        def initialize(url)
+          super <<~TXT
+            Error fetching '#{url}' with Dragonfly!
+
+            This URL is not allowed through the proxy, and therefore the
+            request was blocked by the Workarea Commerce Cloud firewall.
+
+            To allow the host through your proxy server, run the following
+            command on your local development workstation:
+
+                workarea [qa|staging|production] edit proxy
+
+            Thanks for using Commerce Cloud, and keeping your Workarea
+            application safe.
+          TXT
+        end
+      end
+
       def get(url)
         url = parse_url(url)
         proxy_url = ENV['HTTPS_PROXY'] || ENV['HTTP_PROXY']
@@ -15,7 +34,16 @@ module Dragonfly
           request.basic_auth(url.user, url.password)
         end
 
-        http.request(request)
+        response = http.request(request)
+
+        raise ProxyError, url if blocked?(response)
+
+        response
+      end
+
+      def blocked?(response)
+        response.code.to_i == 403 &&
+          response.header['X-Squid-Error'] == 'ERR_ACCESS_DENIED 0'
       end
     end
   end
--- a/core/lib/workarea/ext/freedom_patches/dragonfly_job_fetch_url.rb
+++ b/core/lib/workarea/ext/freedom_patches/dragonfly_job_fetch_url.rb
@@ -7,7 +7,7 @@ module Dragonfly
             Error fetching '#{url}' with Dragonfly!
 
             This URL is not allowed through the proxy, and therefore the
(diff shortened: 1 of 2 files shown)


core/lib/workarea/ext/freedom_patches/dragonfly_job_fetch_url.rb
```diff
@@ -1,6 +1,16 @@
 module Dragonfly
   class Job
     class FetchUrl < Step
+      class ProxyError < StandardError
+        def initialize(url)
+          super %(
+            Error fetching #{url}:
```
tubbo, line 7: open to any suggestions on the text here, I just kinda wrote this out as a first draft

core/lib/workarea/ext/freedom_patches/dragonfly_job_fetch_url.rb
```diff
@@ -15,7 +34,16 @@ def get(url)
...
 
-        http.request(request)
+        response = http.request(request)
+
+        raise ProxyError, url if blocked?(response)
+
+        response
+      end
+
+      def blocked?(response)
```
tubbo, line 44: talking with @MMartyn, the proxy should throw back a 403 at us and this special header rather than time out. a timeout would be the result of _not_ having a proxy configured at all, which shouldn't be the case here so long as you're using our hosting.

core/test/lib/workarea/ext/freedom_patches/dragonfly_job_fetch_url_test.rb
```diff
@@ -0,0 +1,24 @@
+require 'test_helper'
+
+module Workarea
+  class DragonflyJobFetchURLTest < TestCase
+    def test_fail_early_on_misconfigured_proxy
+      product = create_product
+
+      stub_request(:any, 'http://via.placeholder.com/350x150')
```
tubbo, line 8: This is just [Webmock](https://github.com/bblimke/webmock), used directly, in order to stub out the request to `via.placeholder.com`. I couldn't find any 3rd-party services that would return me a custom status code _and_ headers, but I kinda like this solution better anyway as it doesn't rely on anything we weren't depending on before.

core/lib/workarea/ext/freedom_patches/dragonfly_job_fetch_url.rb
```diff
@@ -1,6 +1,25 @@
 module Dragonfly
   class Job
     class FetchUrl < Step
+      class ProxyError < RuntimeError
+        def initialize(url)
+          super <<~TXT
+            Error fetching '#{url}' with Dragonfly!
+
+            This URL is not allowed through the proxy, and therefore the
+            request was blocked by the Workarea Commerce Cloud firewall.
```
bencrouse, line 10: We don't really use the term firewall much, better to use the conventional language.
```suggestion
            request was blocked by the Workarea Commerce Cloud proxy.
```

tubbo:
It should raise a `ProxyError` in the Sidekiq job as the CSV is being imported, so the user would not actually see anything.

As far as moving it to the plugin, I'm into that but I feel like if we're going to move this error over, we should just move the whole freedom patch into the plugin, as it's not relevant for any installation of Workarea that does not require a production server proxy.

bencrouse:
* The original reported problem is "I uploaded my CSV and I have no idea what happened". So we'll want to make sure this displays to the user.
* As far as moving it, I agree sounds good to me. I'll close the PR.

PR #456, opened 2020-06-23
Reset Geocoder between tests

bencrouse:
This ensures individual tests monkeying around with Geocoder config will
get restored before the next test runs.


bencrouse:
Moved here for v3.4 support: https://github.com/workarea-commerce/workarea/pull/458

PR #460, opened 2020-06-25
Bump rack version

bencrouse:
Fixes CVE-2020-8184


tubbo:
Surprised our good friend Dependabot didn’t catch this.

PR #461, opened 2020-06-30
Patch Jbuilder to Support Varying Cache

tubbo:
Previously, admins were not able to see up-to-date data in API requests due to the `#cache!` method in Jbuilder not being patched to skip caching when an admin is logged in. To resolve this, Workarea now applies the same patch to Jbuilder as it does to ActionView. Reading from the cache is now skipped if you're logged in as an admin, and cache keys are appended with the configured `Cache::Varies` just the same as in regular Haml views.

### DIFF
--- a/core/lib/workarea/ext/jbuilder/jbuilder_cache.rb
+++ b/core/lib/workarea/ext/jbuilder/jbuilder_cache.rb
@@ -1,0 +1,29 @@
+decorate JbuilderTemplate, with: :workarea do
+  def _cache_fragment_for(*)
+    return yield if workarea_admin?
+
+    super
+  end
+
+  def _cache_key(*)
+    super.tap do |result|
+      result << workarea_cache_varies if workarea_cache_varies.present?
+    end
+  end
+
+  private
+
+  def workarea_admin?
+    @context&.controller&.current_user&.admin?
+  rescue ::RuntimeError
+    false
+  end
+
+  def workarea_cache_varies
+    workarea_request_env['workarea.cache_varies']
+  end
+
+  def workarea_request_env
+    @context.controller.request.env || {}
+  end
+end
--- a/core/lib/workarea/core.rb
+++ b/core/lib/workarea/core.rb
@@ -139,6 +139,7 @@ require 'workarea/ext/mongoid/audit_log_entry.decorator'
 require 'workarea/ext/mongoid/find_ordered'
 require 'workarea/ext/sprockets/ruby_processor'
 require 'workarea/ext/jbuilder/jbuilder_append_partials'
+require 'workarea/ext/jbuilder/jbuilder_cache'
 
 if Rails.env.development?
   require 'workarea/ext/freedom_patches/routes_reloader'


core/lib/workarea/ext/jbuilder/jbuilder_cache.rb
```diff
@@ -0,0 +1,31 @@
...
+  rescue ::RuntimeError
+    false
+  end
+
+  def workarea_cache_varies
+    workarea_request_env['workarea.cache_varies']
+  end
+
+  def workarea_request_env
+    @context.controller.request.env || {}
```
tubbo, line 29: This `env` isn't necessarily set in places like helper tests, or when rendering a JSON response from a controller programmatically (using `ActionController::Base.render`), so we gotta be extra safe here.

core/lib/workarea/ext/jbuilder/jbuilder_cache.rb
```diff
@@ -0,0 +1,31 @@
...
+    super.tap do |result|
+      result << workarea_cache_varies if workarea_cache_varies.present?
+    end
+  end
+
+  private
+
+  def workarea_admin?
+    @context&.controller&.current_user&.admin?
+  rescue ::RuntimeError
```
tubbo, line 20: This should only really happen in the storefront API, but we don't know whether API is installed yet so I'm rescuing the base `RuntimeError` class here.

core/lib/workarea/ext/jbuilder/jbuilder_cache.rb
```diff
@@ -0,0 +1,31 @@
+require 'jbuilder/jbuilder_template'
+
+decorate JbuilderTemplate do
```
tubbo, line 3: I had problems when just overriding the class because `JbuilderTemplate` has some absurd way it's getting loaded into the codebase. By making this a decorator, we're ensuring that it will get applied after the class has already been loaded and defined, so it will be possible to call `super` within it. Previously, I basically couldn't call `super` without it erroring.

core/lib/workarea/ext/jbuilder/jbuilder_cache.rb
```diff
@@ -0,0 +1,31 @@
+require 'jbuilder/jbuilder_template'
```
bencrouse, line 1: Was this necessary? I'd think the jbuilder require in core.rb would have taken care of it.

core/lib/workarea/ext/jbuilder/jbuilder_cache.rb
```diff
@@ -0,0 +1,31 @@
...
+
+decorate JbuilderTemplate, with: :workarea do
+  def _cache_fragment_for(key, options)
```
bencrouse, line 4: If this isn't going to use args, better to use `_cache_fragment_for(*)` so if anything about the args changes we don't have to update.

core/lib/workarea/ext/jbuilder/jbuilder_cache.rb
```diff
@@ -0,0 +1,31 @@
...
+    return yield if workarea_admin?
+
+    super
+  end
+
+  def _cache_key(key, options)
```
bencrouse, line 10: Same thing with args here.

core/lib/workarea/ext/jbuilder/jbuilder_cache.rb
tubbo: Whoops, yeah I think I was trying something out here.

PR #463, opened 2020-07-10
Improve Content Area Select UX

tubbo:
Remove the current content name and replace it with a static label indicating what the `<select>` menu to the right of it is selecting, which is the current content area. This UI only displays when there are multiple areas for a given `Content`.

### DIFF
--- a/admin/app/views/workarea/admin/content/_edit.html.haml
+++ b/admin/app/views/workarea/admin/content/_edit.html.haml
@@ -7,7 +7,7 @@
       .grid.grid--middle
         .grid__cell.grid__cell--50
           - if content.show_areas?
-            %span= content.name
+            %span= t('workarea.admin.content.edit.current_area')
             = inline_svg('workarea/admin/icons/navigate_next.svg', class: 'svg-icon svg-icon--small svg-icon--white')
             = form_tag request.path, method: 'get', class: 'content-editor__area-select-form' do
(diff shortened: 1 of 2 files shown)


admin/config/locales/en.yml
```diff
@@ -664,6 +664,7 @@ en:
           loading: Loading...
           preview: Preview
           title: Content for %{content}
+          current_area: 'Editing Content Area:'
```
bencrouse, line 667:
```suggestion
          current_area: 'Current Area:'
```

admin/app/views/workarea/admin/content/_edit.html.haml
```diff
@@ -7,7 +7,7 @@
       .grid.grid--middle
         .grid__cell.grid__cell--50
           - if content.show_areas?
-            %span= content.name
+            %span= t('workarea.admin.content.edit.current_area')
```
bencrouse, line 10:
```suggestion
            %span #{t('workarea.admin.content.edit.current_area')}:
```

admin/config/locales/en.yml
```diff
@@ -664,6 +664,7 @@ en:
           loading: Loading...
           preview: Preview
           title: Content for %{content}
+          current_area: 'Current Area:'
```
bencrouse, line 667:
```suggestion
          current_area: 'Current Area'
```

PR #464, opened 2020-07-10
Configure Sliced Credit Card Attributes

tubbo:
To prevent an unnecessary decoration of the `Payment` class, make the attributes used in `Payment#set_credit_card` configurable. This configuration is reused on the `Users::CreditCardsController` for adding a new saved credit card to a user's account.


tubbo:
@bencrouse I tried to use the config there but it's slightly different attributes, particularly the controller permits the additional `:first_name` and `:last_name` attributes which the `Payment` doesn't necessarily need. So it's not really a 1:1 mapping.

PR #468, opened 2020-07-15
Update change-product-placeholder-image.html.md

MMartyn:
Add caveat that an application restart may be required to see the new placeholder image.


bencrouse:
I cherry picked this into `v3.5-stable`

PR #469, opened 2020-07-15
Add admin configuration fields for safe/block listing IP addresses

mttdffy:
WORKAREA-247

### DIFF
--- a/core/config/initializers/00_configuration.rb
+++ b/core/config/initializers/00_configuration.rb
@@ -179,4 +179,18 @@ Workarea::Configuration.define_fields do
       },
       description: 'Subjects list for the contact form'
   end
+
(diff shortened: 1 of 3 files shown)


core/config/initializers/00_configuration.rb
```diff
@@ -179,4 +179,18 @@
       },
       description: 'Subjects list for the contact form'
   end
+
+  fieldset 'Security', namespaced: false do
+    field 'Safe IP Addresses',
+      id: :safe_ip_addresses,
+      type: :array,
+      default: [],
+      description: 'List of known IP addresses that should always have access to the site.'
```
bencrouse, line 188: Would you change the language or add a tooltip or something to clarify that they can specify range with the mask?

PR #472, opened 2020-07-20
Add js module to allow inserting remote requests onto the page

mttdffy:


admin/app/assets/javascripts/jquery_ujs/insert.js
```diff
@@ -0,0 +1,22 @@
...
+        $target = $($element.data('insert')),
+        $content = $(response);
+
+    if (_.isEmpty($target)) { return; }
+    if (! $element.is('[data-insert]')) { return; }
+
+    if ($element.is('[data-insert-after]')) {
+        $content.insertAfter($target);
+    } else {
+        $target.html($content);
```
bencrouse, line 18: Since this is called "insert", it seems a little weird to use `html` since that will replace the content of `$target`. Did you mean `append`?

mttdffy: I didn't, but I also realized i don't need the `html` behavior in product bundles anymore. So I just update the module

PR #474, opened 2020-07-27
Configurable Order Item Equivalence Attributes

tubbo:
When adding an item that already exists in the cart, Workarea previously checked the SKU and customizations on the item before deciding whether to actually build a new `Order::Item` or update the one found in place.  This is somewhat limiting for plugins such as `Workarea::Subscriptions`, which add more data to an order item that is expected to be updated in place (so long as the item is a subscription). To address this, a configuration setting has been added which plugins and apps can append to in order to affect the "uniqueness" of a particular order item. By appending to this config, both host apps and plugins can tell the `Order::ItemsExtension#find_existing` method to factor in additional fields on the order when determining whether a given item exists in the cart or not.

### DIFF
--- a/core/app/models/workarea/order.rb
+++ b/core/app/models/workarea/order.rb
@@ -242,9 +242,8 @@ module Workarea
     def add_item(attributes)
       quantity = attributes.fetch(:quantity, 1).to_i
       sku = attributes[:sku]
-      customizations = attributes[:customizations]
 
-      if existing_item = items.find_existing(sku, customizations)
+      if existing_item = items.find_existing(sku, attributes)
         update_item(existing_item.id, quantity: existing_item.quantity + quantity)
       else
         items.build(attributes)
@@ -253,15 +252,18 @@ module Workarea
       save
     end
 
-    # Updates an items attributes
+    # Update an item's attributes. When quantity is provided for an
+    # existing item, increase quantity by the provided value rather than
+    # set the quantity to the passed-in value.
     #
     # @param [String] id
-    # @param [Hash] attributes new item attributes
+    # @param [Hash] attributes - new item attributes
     #
     # @return [Boolean]
     #   whether the item was successfully updated
     def update_item(id, attributes)
-      existing_item = items.find_existing(attributes[:sku], attributes[:customizations])
+      existing_item = items.find_existing(attributes[:sku], attributes)
+
       if existing_item.present? && existing_item.id.to_s != id.to_s
         item = items.find(id)
         existing_item.update_attributes(quantity: existing_item.quantity + (attributes[:quantity] || item.quantity))
--- a/core/app/models/workarea/order/item.rb
+++ b/core/app/models/workarea/order/item.rb
@@ -156,9 +156,9 @@ module Workarea
     # @return [Boolean] whether this item's attributes is equal to the
     #                   passed-in parameters.
     def attributes_eql?(params = {})
-      equivs = Workarea.config.distinct_order_item_attributes.map(&:to_sym)
-      attrs = params&.symbolize_keys&.slice(*equivs) || {}
-      customizations = params[:customizations]&.stringify_keys!
+      params = params.to_h.with_indifferent_access
+      attrs = params.slice(*Workarea.config.distinct_order_item_attributes)
+      customizations = params[:customizations]&.stringify_keys
       distinct = attrs.all? do |attribute, value|
         self[attribute] == value
       end
--- a/core/app/models/workarea/order/item.rb
+++ b/core/app/models/workarea/order/item.rb
@@ -147,13 +147,32 @@ module Workarea
       customizations.present?
     end
 
+    # Determine whether the additional fields on this item are
+    # equivalent to the passed-in Hash of parameters. Uses the
+    # `distinct_order_item_attributes` configuration setting to filter
+    # down which parameters it will be checking equivalence on.
+    #
+    # @param [Hash] params - Updated order item parameters.
+    # @return [Boolean] whether this item's attributes is equal to the
+    #                   passed-in parameters.
+    def attributes_eql?(params = {})
+      equivs = Workarea.config.distinct_order_item_attributes.map(&:to_sym)
+      attrs = params&.symbolize_keys&.slice(*equivs) || {}
+      customizations = params[:customizations]&.stringify_keys!
+      distinct = attrs.all? do |attribute, value|
+        self[attribute] == value
+      end
+
+      distinct && customizations_eql?(customizations)
+    end
+
     # Determine whether the customizations of this item
     # are equivalent to the customizations of another.
     # This is used when updating/adding items so we can
     # see whether we should merge items that have the
     # same SKU but different customizations.
     #
-    # @param test [Order::Item]
+    # @param [Hash] test
     # @return [Boolean]
     #
     def customizations_eql?(test)
--- a/core/app/models/workarea/order/items_extension.rb
+++ b/core/app/models/workarea/order/items_extension.rb
@@ -1,13 +1,8 @@
 module Workarea
   class Order
     module ItemsExtension
-      def find_existing(sku, customizations = {})
-        customizations.stringify_keys! if customizations.present?
-
-        detect do |item|
-          item.sku == sku &&
-            item.customizations_eql?(customizations)
-        end
+      def find_existing(sku, attributes = {})
+        detect { |item| item.sku == sku && item.attributes_eql?(attributes) }
       end
     end
   end
--- a/core/app/services/workarea/add_multiple_cart_items/item.rb
+++ b/core/app/services/workarea/add_multiple_cart_items/item.rb
@@ -23,7 +23,7 @@ module Workarea
         return @item&.persisted? if defined?(@item)
 
         order.add_item(item_params).tap do |result|
-          @item = order.items.find_existing(sku, customizations&.to_h) if result
+          @item = order.items.find_existing(sku, item_params) if result
         end
       end
 
--- a/core/lib/workarea/configuration.rb
+++ b/core/lib/workarea/configuration.rb
@@ -1321,6 +1321,16 @@ module Workarea
         cvv
         amount
       ]
+
+      # This is the Array of Symbols used to determine whether
+      # `Order#add_item` will create a new `Order::Item` record or
+      # whether it will update an existing one. The values of this Array
+      # are `Order::Item` field names, used to distinguish items from
+      # one another in a single `Order`. In other words, when these
+      # attributes differ between an existing item and a newly added
+      # item to the cart, the newly added item will appear separately in
+      # the Order.
+      config.distinct_order_item_attributes = []
     end
   end
 end
--- a/core/test/models/workarea/order_test.rb
+++ b/core/test/models/workarea/order_test.rb
@@ -85,6 +85,34 @@ module Workarea
       order.add_item(product_id: '1234', sku: 'SKU', quantity: 2)
       assert_equal(1, order.items.count)
       assert_equal(4, order.items.last.quantity)
+
+      assert_no_changes 'order.items.count' do
+        order.add_item(
+          product_id: '1234',
+          sku: 'SKU',
+          quantity: 1
+        )
+      end
+
+      Workarea.config.distinct_order_item_attributes << :discountable
+
+      assert_changes 'order.items.count' do
+        order.add_item(
+          product_id: '1234',
+          sku: 'SKU',
+          quantity: 1,
+          discountable: false
+        )
+      end
+
+      assert_no_changes 'order.items.count' do
+        order.add_item(
+          product_id: '1234',
+          sku: 'SKU',
+          quantity: 1,
+          discountable: true
+        )
+      end
     end
 
     def test_update_item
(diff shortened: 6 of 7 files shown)


core/app/models/workarea/order/item.rb
```diff
@@ -166,5 +166,22 @@ def customizations_eql?(test)
...
+    # down which parameters it will be checking equivalence on.
+    #
+    # @param [Hash] params - Updated order item parameters.
+    # @return [Boolean] whether this item's attributes is equal to the
+    #                   passed-in parameters.
+    def attributes_eql?(params = {})
+      equivs = Workarea.config.distinct_order_item_attributes.map(&:to_sym)
+      attrs = params&.symbolize_keys&.slice(*equivs)
+
+      equivs.blank? || params.blank? || attrs.any? do |attribute, value|
```
mttdffy, line 182: shouldn't this be `all?`

core/app/models/workarea/order.rb
```diff
@@ -242,29 +242,35 @@ def place
...
       quantity = attributes.fetch(:quantity, 1).to_i
       sku = attributes[:sku]
-      customizations = attributes[:customizations]
 
-      if existing_item = items.find_existing(sku, customizations)
-        update_item(existing_item.id, quantity: existing_item.quantity + quantity)
+      if existing_item = items.find_existing(sku, attributes)
+        attributes[:quantity] = existing_item.quantity + quantity
+
+        update_item(existing_item.id, attributes.except(:via))
```
mttdffy, line 249: still think quantity should be the only thing pass through here.

tubbo: we did talk about that, yeah. dunno what i was thinking.

core/app/models/workarea/order/item.rb
tubbo: yup.

core/app/models/workarea/order.rb
```diff
@@ -253,18 +252,23 @@ def add_item(attributes)
       save
     end
 
-    # Updates an items attributes
+    # Update an item's attributes. Increases quantity by the provided
```
tubbo, line 255: This method didn't change in its feature set, but the docs were just lacking a bit.

core/app/models/workarea/order/item.rb
```diff
@@ -153,7 +153,7 @@ def customized?
     # see whether we should merge items that have the
     # same SKU but different customizations.
     #
-    # @param test [Order::Item]
+    # @param [Hash] test
```
tubbo, line 156: This seemed to be incorrect.

core/app/models/workarea/order.rb
```diff
@@ -253,18 +252,23 @@ def add_item(attributes)
...
     # @return [Boolean]
     #   whether the item was successfully updated
     def update_item(id, attributes)
-      existing_item = items.find_existing(attributes[:sku], attributes[:customizations])
+      existing_item = items.find_existing(attributes[:sku], attributes)
+
       if existing_item.present? && existing_item.id.to_s != id.to_s
         item = items.find(id)
-        existing_item.update_attributes(quantity: existing_item.quantity + (attributes[:quantity] || item.quantity))
+        attributes[:quantity] = existing_item.quantity + (
```
bencrouse, line 268: Changing `attributes` here is unnecessary mutation of the hash that's been passed in, let's leave this as it was.

core/app/models/workarea/order.rb
```diff
@@ -242,9 +242,8 @@ def place
     def add_item(attributes)
       quantity = attributes.fetch(:quantity, 1).to_i
       sku = attributes[:sku]
-      customizations = attributes[:customizations]
 
-      if existing_item = items.find_existing(sku, customizations)
+      if existing_item = items.find_existing(sku, attributes)
```
bencrouse, line 246: Why not just pass the whole hash here?
```suggestion
      if existing_item = items.find_existing(attributes)
```

core/app/models/workarea/order/item.rb
```diff
@@ -166,5 +166,22 @@ def customizations_eql?(test)
...
+    # down which parameters it will be checking equivalence on.
+    #
+    # @param [Hash] params - Updated order item parameters.
+    # @return [Boolean] whether this item's attributes is equal to the
+    #                   passed-in parameters.
+    def attributes_eql?(params = {})
+      equivs = Workarea.config.distinct_order_item_attributes.map(&:to_sym)
+      attrs = params&.symbolize_keys&.slice(*equivs)
+
+      equivs.blank? || params.blank? || attrs.all? do |attribute, value|
```
bencrouse, line 182: I think this should be false if the configuration is blank - by setting those blank you're saying "there are no attributes that make 2 order items equivalent".

Also `all?` will return `true` on an empty enumerable so this logic needs some clean up.

core/app/models/workarea/order/items_extension.rb
```diff
@@ -1,12 +1,14 @@
...
     module ItemsExtension
-      def find_existing(sku, customizations = {})
+      def find_existing(sku, attributes = {})
+        customizations = attributes[:customizations] if attributes.present?
         customizations.stringify_keys! if customizations.present?
 
         detect do |item|
           item.sku == sku &&
-            item.customizations_eql?(customizations)
+            item.customizations_eql?(customizations) &&
```
bencrouse, line 10: We should include `customizations_eql?` as part of the check, it would be unexpected to check `attributes_eql?` and get back `true` if the customizations were actually different.

core/app/models/workarea/order/item.rb
```diff
@@ -166,5 +166,22 @@ def customizations_eql?(test)
...
+
+    # Determine whether the additional fields on this item are
+    # equivalent to the passed-in Hash of parameters. Uses the
+    # `distinct_order_item_attributes` configuration setting to filter
+    # down which parameters it will be checking equivalence on.
+    #
+    # @param [Hash] params - Updated order item parameters.
+    # @return [Boolean] whether this item's attributes is equal to the
+    #                   passed-in parameters.
+    def attributes_eql?(params = {})
```
bencrouse, line 178: This name is a little unclear, how can we name it represent that this is different from `order_item.attributes == other_order_item.attributes`

core/app/services/workarea/add_multiple_cart_items/item.rb
```diff
@@ -23,7 +23,7 @@ def save
         return @item&.persisted? if defined?(@item)
 
         order.add_item(item_params).tap do |result|
-          @item = order.items.find_existing(sku, customizations&.to_h) if result
+          @item = order.items.find_existing(sku, item_params) if result
```
bencrouse, line 26: Anyone know how many plugins use `find_existing`?

core/lib/workarea/configuration.rb
```diff
@@ -1321,6 +1321,13 @@ def self.setup_defaults
         cvv
         amount
       ]
+
+      # This is the Array of Symbols used to determine whether
+      # `Order#add_item` will create a new `Order::Item` record or
+      # whether it will update an existing one. Attributes from
+      # `Catalog::Customizations` are always included in this existing
```
bencrouse, line 1328: As written, these docs aren't true, customizations are handled in a separate method. Let's make this description true.

core/app/models/workarea/order/items_extension.rb
tubbo: We don't check `attributes_eql?` until `customizations_eql?` is true. To preserve the functionality from v3.5 and below, we're only going to check the attributes if the SKU and customization check fails.

core/app/models/workarea/order/item.rb
tubbo: Agreed, not really sure what to call it though. I used this method name as a stand-in for a future name.

core/app/services/workarea/add_multiple_cart_items/item.rb
tubbo: https://stash.tools.weblinc.com/plugins/servlet/search?q=find_existing

core/lib/workarea/configuration.rb
tubbo: Ah yeah, I wanted to be clear that customizations are always checked for equivalence in a separate method, but I wasn't sure how to say that here.

core/app/models/workarea/order.rb
tubbo: yeah originally i had `dup`-ed it but then i realized we probably don't want to update attrs in `update_item`.

core/app/models/workarea/order.rb
tubbo: I felt that changing the arity of the method signature here would be an unnecessary breaking change.

core/app/services/workarea/add_multiple_cart_items/item.rb
tubbo: I found this one use in BOPUS: https://github.com/workarea-commerce/workarea-bopus/blob/64df917e175e3c584ad316a2a83ef3eeeee91103/app/models/workarea/order.decorator

And this one in Channel Advisor: https://stash.tools.weblinc.com/plugins/servlet/search?q=project%3AWP%20find_existing

core/app/models/workarea/order/items_extension.rb
bencrouse: I'm not sure I follow, `attributes_eql?` is really able to replace all the logic in this `detect` block as a point to override and customize that logic without having to get at this awkwardly deportable method.

core/app/models/workarea/order/item.rb
bencrouse: Time to dance the side convo shuffle.

core/app/services/workarea/add_multiple_cart_items/item.rb
bencrouse: Ok, I guess not too bad.

core/app/models/workarea/order.rb
mttdffy: This comment is misleading. Quantity is only increased by the value passed, rather than set to that value, if you pass an id of an item that doesn't match the item found by `find_existing`, as its essentially merging items at that point. Under normal conditions `update_item` sets the value of the item matching the id passed to the values passed.

core/app/models/workarea/order/item.rb
tubbo: > by setting those blank you're saying "there are no attributes that make 2 order items equivalent".

That's what preserves the original functionality, because before this code was introduced, there _were_ no attributes that made two order items equivalent, other than customizations, which if `customizations_eql?` returns true this method isn't even called.

tubbo:
@mttdffy I'm down for that, the ItemsExtension thing has always bugged me. I remember having to/wanting to decorate that in the past and finding it quite impossible.

core/app/models/workarea/order/item.rb
```diff
@@ -147,13 +147,32 @@ def customized?
...
+      equivs = Workarea.config.distinct_order_item_attributes.map(&:to_sym)
+      attrs = params&.symbolize_keys&.slice(*equivs)
+      customizations = params[:customizations]&.stringify_keys! || {}
+      distinct = attrs.all? do |attribute, value|
```
bencrouse, line 162: As written, `attrs` could be `nil` because of the safe operator stuff. In that case, calling `all?` on it here will raise.

core/app/models/workarea/order/item.rb
```diff
@@ -147,13 +147,32 @@ def customized?
...
+    # `distinct_order_item_attributes` configuration setting to filter
+    # down which parameters it will be checking equivalence on.
+    #
+    # @param [Hash] params - Updated order item parameters.
+    # @return [Boolean] whether this item's attributes is equal to the
+    #                   passed-in parameters.
+    def attributes_eql?(params = {})
+      equivs = Workarea.config.distinct_order_item_attributes.map(&:to_sym)
+      attrs = params&.symbolize_keys&.slice(*equivs)
+      customizations = params[:customizations]&.stringify_keys! || {}
```
bencrouse, line 161: `customizations_eql?` handles the blank case, so this doesn't need a default value.

core/app/models/workarea/order.rb
tubbo: I'll clarify that part, thanks.

core/app/models/workarea/order/item.rb
tubbo: ah, good catch...I think I meant to do that `|| {}` on the line above.

core/app/models/workarea/order/item.rb
```diff
@@ -147,13 +147,32 @@ def customized?
...
+    # `distinct_order_item_attributes` configuration setting to filter
+    # down which parameters it will be checking equivalence on.
+    #
+    # @param [Hash] params - Updated order item parameters.
+    # @return [Boolean] whether this item's attributes is equal to the
+    #                   passed-in parameters.
+    def attributes_eql?(params = {})
+      equivs = Workarea.config.distinct_order_item_attributes.map(&:to_sym)
+      attrs = params&.symbolize_keys&.slice(*equivs) || {}
+      customizations = params[:customizations]&.stringify_keys!
```
bencrouse, line 161: So in reading this closer, two things:
*  `params` being `nil` is handled on line 160, but not 161, so that's inconsistent. 
* `params[:customizations]` is getting mutated with `stringify_keys!` and it's always a good idea to avoid mutating something passed in if reasonable.

A cleaner approach handling this whole situation could be:

```suggestion
    def attributes_eql?(params = {})
      params = params.to_h.with_indifferent_access
      attrs = params.slice(*Workarea.config.distinct_order_item_attributes)
      customizations = params[:customizations]&.stringify_keys
```

PR #475, opened 2020-07-27
Fix Clone URL in CLI Documentation

tubbo:
The clone URL for the `homebrew-taps` repo on Stash required SSH access to the repo, whereas read-only members may not have that level of access. Update this to include a more accessible URL.

Kinda weird that we don't ever indicate that you actually need to be on Commerce Cloud to use this thing, but maybe that's a discussion we should have down the road...


bencrouse:
How does the authentication work for the `https` version?

PR #477, opened 2020-07-28
Allow inquiry subjects to be localized

mttdffy:
WORKAREA-238


mttdffy:
This is definitely a band-aid, but it accomplishes the task for now. Ideally, maybe in 3.6, we would enable configurable fields to be localized.

mttdffy:
it would work for v3.4 but makes less sense given its a static config in 3.4 and they can already just change it how they want.

PR #481, opened 2020-08-04
Return Status of `Checkout#update`

tubbo:
For APIs and other consumers of the Checkout model, return a boolean response from the `#update` method to signify whether the operation succeeded or failed. This response is used directly in the API to return an `:unprocessable_entity` response code when an update operation fails.

### DIFF
--- a/core/app/models/workarea/checkout.rb
+++ b/core/app/models/workarea/checkout.rb
@@ -158,10 +158,14 @@ module Workarea
     # Used in auto completing an order for a logged in user.
     #
     # @param [Hash] parameters for updating
-    # @return [self]
+    # @return [Boolean] whether the update was successful.
     #
     def update(params = {})
-      steps.each { |s| s.new(self).update(params) }
+      return true if params.blank?
+
+      steps.reduce(true) do |result, step|
+        result &= step.new(self).update(params)
+      end
     end
 
     # Whether this checkout needs any further information
--- a/core/test/models/workarea/checkout_test.rb
+++ b/core/test/models/workarea/checkout_test.rb
@@ -51,6 +51,63 @@ module Workarea
       refute_equal(payment_id, checkout.payment.object_id)
     end
 
+    def test_update
+      create_shipping_service
+      checkout = Checkout.new(@order)
+
+      checkout.start_as(:guest)
+
+      refute(
+        checkout.update(
+          email: 'test@workarea.com',
+          shipping_address: {
+            last_name: 'Crouse',
+            street: '22 S. 3rd St.',
+            street_2: 'Second Floor',
+            city: 'Philadelphia',
+            region: 'PA',
+            postal_code: '19106',
+            country: 'US'
+          },
+          billing_address: {
+            first_name: 'Ben',
+            street: '22 S. 3rd St.',
+            street_2: 'Second Floor',
+            city: 'Philadelphia',
+            region: 'PA',
+            postal_code: '19106',
+            country: 'US'
+          }
+        )
+      )
+
+      assert(
+        checkout.update(
+          email: 'test@workarea.com',
+          shipping_address: {
+            first_name: 'Ben',
+            last_name: 'Crouse',
+            street: '22 S. 3rd St.',
+            street_2: 'Second Floor',
+            city: 'Philadelphia',
+            region: 'PA',
+            postal_code: '19106',
+            country: 'US'
+          },
+          billing_address: {
+            first_name: 'Ben',
+            last_name: 'Crouse',
+            street: '22 S. 3rd St.',
+            street_2: 'Second Floor',
+            city: 'Philadelphia',
+            region: 'PA',
+            postal_code: '19106',
+            country: 'US'
+          }
+        )
+      )
+    end
+
     def test_continue_as
       checkout = Checkout.new(@order)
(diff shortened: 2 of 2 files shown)


core/app/models/workarea/checkout.rb
```diff
@@ -161,7 +161,13 @@ def continue_as(user)
     # @return [self]
     #
     def update(params = {})
-      steps.each { |s| s.new(self).update(params) }
+      steps.all? do |step|
+        result = step.new(self).update(params)
+
+        next true if result.nil?
```
tubbo, line 167: this right here allows for steps for which we aren't passing params to be "skipped" when determining whether the update failed or not. otherwise, it would probably just fail every time since `#update` would be returning `nil` for many of these steps.

core/app/models/workarea/checkout.rb
```diff
@@ -161,7 +161,14 @@ def continue_as(user)
...
     #
     def update(params = {})
-      steps.each { |s| s.new(self).update(params) }
+      steps.all? do |step_class|
+        step = step_class.new(self)
+        result = step.update(params)
+
+        next true if step.complete?
+
+        result
```
mttdffy, line 170: does it ever make sense to return result? Wondering if this should just be

```
steps.all? do |step_class|
  step = step_class.new(self)
  step.update(params)
  step.complete?
end
```

tubbo:
this build may not ever end :D

core/app/models/workarea/checkout.rb
tubbo: yup, that's much more concise and does the same thing

tubbo:
@mttdffy Yeah, that seems to be the issue I'm running into now. When you're mid-checkout, obviously not all steps are going to be `#complete?`, so it always fails even though all of the steps you've encountered are complete. I'm thinking `next true if step.update(params).nil?` would fix that. In other words, we still need to check whether `#update` just returned out early vs tried to update and failed.

This might be too big for a patch release, but we should definitely consider refactoring how this works someday. If there was a central `#errors` object on `Workarea::Checkout` that mimicked an `ActiveModel::Errors`, we could just check `errors.any?` at the end of this method.

**edit:** apparently, I was wrong about my assumption :( https://github.com/workarea-commerce/workarea/blob/master/core/app/models/workarea/checkout/steps/payment.rb#L19

core/app/models/workarea/checkout.rb
```diff
@@ -161,7 +161,9 @@ def continue_as(user)
     # @return [self]
     #
     def update(params = {})
-      steps.each { |s| s.new(self).update(params) }
+      steps.reduce(true) do |_, step|
+        step.new(self).update(params)
```
bencrouse, line 165: This will only ever return the result of the last step.

core/app/models/workarea/checkout.rb
```diff
@@ -161,7 +161,9 @@ def continue_as(user)
     # @return [self]
     #
     def update(params = {})
```
bencrouse, line 163: We discussed a `return true if params.blank?` to mimic calling `update` on a model where nothing has changed.

core/app/models/workarea/checkout.rb
tubbo: Yeah I'm sorta confused about the whole way this works. When I print out all the update results, I get `false` for almost every one and then `true` for payment. But if I add `result && step.new(self).update(params)`, my assertion for whether checkout updates does not pass, because `result` is going to be false, so it never actually runs update for any of the steps.

core/app/models/workarea/checkout.rb
```diff
@@ -161,8 +161,10 @@ def continue_as(user)
...
-      steps.reduce(true) do |_, step|
-        step.new(self).update(params)
+      return true if params.blank?
+
+      steps.reduce(true) do |result, step|
+        result && step.new(self).update(params)
```
tubbo, line 167: This breaks the test I wrote, in that `#update` never returns `true` and won't even get to the additional steps with this code added in. I wanted to showcase it here so we can start discussing a way around the problem. Part of this is that the return values of these methods were never really considered before, so we're shoehorning in this new functionality without really taking a step back and seeing if this is the right approach. I personally have and always will advocate for a separate system for bubbling up and tracking errors in the checkout, such as an ActiveModel::Errors object that just collects errors from all of the models in checkout. If we did things this way, the `#update` method could return back to its original form (by just running an `#each` loop), then we can just check `errors.any?` at the end to determine whether the update succeeded. No errors = update success. Let's stop messing around with this stupid boolean logic and just fix the problem already.

bencrouse: Hmm, the struggle is real. Boolean short-circuiting might be an issue as-written where it won't bother to run the second part of the `&&` when the first part is already `false`.

Maybe
```
result = true
steps.each { |s| result &= s.new(self).update(params) }
result
```

`&=` won't short circuit the way `&&` will:
```
irb(main):013:0> x = Proc.new { puts "HI"; true; }
=> #<Proc:0x00007ff59da70518@(irb):13>
irb(main):014:0> y = false
=> false
irb(main):015:0> y && x.call
=> false
irb(main):016:0> y &= x.call
HI
=> false
```

tubbo: that worked!

PR #484, opened 2020-08-11
Remove port from host configuration in installer

bencrouse:
Ports aren't part of hosts, this causes problems when the value is used like a true host.


bencrouse:
@mttdffy I fixed the mailers, what redirects are you talking about?

PR #485, opened 2020-08-13
Fix missing release changes for CSV importing with embedded models

bencrouse:
Trying to update an embedded model via CSV import with a release causes
an existing changeset for the root model to get destroyed. This happens
because the CSV import calls `#save` on the root, which has no changes
so it removes the changeset.

This patch fixes by iterating over the models the CSV row might affect
and calling `#save` on the embedded ones first (if necessary) to ensure
the changesets get correctly created and to avoid calling the save on
the root without changes which removes the existing changeset.

### DIFF
--- a/core/app/models/workarea/data_file/csv.rb
+++ b/core/app/models/workarea/data_file/csv.rb
@@ -17,7 +17,15 @@ module Workarea
           assign_attributes(root, attrs)
           assign_embedded_attributes(root, attrs)
 
-          if root.save || failed_new_record_ids.exclude?(id)
+          possibly_affected_models = root.embedded_children + [root]
+          was_successful = true
+
+          possibly_affected_models.each do |model|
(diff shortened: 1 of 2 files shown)


core/app/models/workarea/data_file/csv.rb
```diff
@@ -17,7 +17,15 @@ def import!
           assign_attributes(root, attrs)
           assign_embedded_attributes(root, attrs)
 
-          if root.save || failed_new_record_ids.exclude?(id)
+          possibly_affected_models = root.embedded_children + [root]
+          was_successful = true
+
+          possibly_affected_models.each do |model|
+            meaningful_changes = model.changes.except('updated_at')
+            was_successful &= model.save if model.changed? && meaningful_changes.present?
```
tubbo, line 25: [hello `&=` my old friend](https://www.youtube.com/watch?v=4fWyzwo1xg0)

PR #488, opened 2020-08-24
Show User Login Tracking Info in Admin

tubbo:
Display the `last_login_at` and `last_login_attempted_at` properties of a User on the attributes card in admin.

### DIFF
--- a/admin/app/views/workarea/admin/users/_cards.html.haml
+++ b/admin/app/views/workarea/admin/users/_cards.html.haml
@@ -35,6 +35,12 @@
                 = link_to model.created_by.name, user_path(model.created_by)
               - else
                 = t('workarea.admin.users.cards.not_applicable')
+            %li
+              %strong= t('workarea.admin.fields.last_login_at')
+              = model.last_login
+            %li
+              %strong= t('workarea.admin.fields.last_login_attempt_at')
+              = model.last_login_attempt
 
             - unless model.admin?
               %li
--- a/admin/app/views/workarea/admin/users/edit.html.haml
+++ b/admin/app/views/workarea/admin/users/edit.html.haml
@@ -80,11 +80,23 @@
             .property
               = label_tag 'user_created_at', t('workarea.admin.fields.created_at'), class: 'property__name'
               = local_time(@user.created_at)
+          - if @user.last_login_at?
+            .grid__cell.grid__cell--50.grid__cell--25-at-medium
+              .property
+                = label_tag 'user_last_login_at', t('workarea.admin.fields.last_login_at'), class: 'property__name'
+                = local_time(@user.last_login_at)
+          - if @user.last_login_attempt_at?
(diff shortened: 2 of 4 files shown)


admin/app/views/workarea/admin/users/_cards.html.haml
```diff
@@ -35,6 +35,12 @@
                 = link_to model.created_by.name, user_path(model.created_by)
               - else
                 = t('workarea.admin.users.cards.not_applicable')
+            %li
+              %strong= t('workarea.admin.fields.last_logged_in_at')
+              = model.last_login
```
tubbo, line 40: This method already existed, but wasn't in use.

admin/config/locales/en.yml
```diff
@@ -1718,6 +1718,8 @@ en:
         last_name: Last Name
         last_order: Last Order
         last_order_at: Last Order At
+        last_logged_in_at: Last Logged In At
```
tubbo, line 1721: dunno if this should really have the "At" on the end, might read a little weird..

admin/app/views/workarea/admin/users/_cards.html.haml
mttdffy: That method is using `time_ago_in_words` in a view model, which I think that method is not defined in and probably throwing an error to cause the test in the build to fail.

admin/config/locales/en.yml
mttdffy: I'd say drop the `at`

admin/app/views/workarea/admin/users/_cards.html.haml
mttdffy: err, ok it includes the date helper.. so maybe its not that. Build is failing though

tubbo: should be resolved once https://github.com/workarea-commerce/workarea/pull/490 is merged.

bencrouse: Yeah, that's the same failure that I was seeing locally.

tubbo:
I fixed the test you were talking about @bencrouse but I think this final failure has to do with impersonation, so i think a mergedown will greenify the build.

admin/app/views/workarea/admin/users/edit.html.haml
```diff
@@ -80,12 +80,24 @@
...
+                = label_tag 'user_last_login_attempt_at', t('workarea.admin.fields.last_login_attempt_at'), class: 'property__name'
+                = local_time(@user.last_login_attempt_at)
+
           - if @user.created_by.present?
             .grid__cell.grid__cell--50.grid__cell--25-at-medium
               .property
                 = label_tag 'user_created_by', t('workarea.admin.fields.created_by'), class: 'property__name'
                 = link_to @user.created_by.name, user_path(@user.created_by)
 
+
```
bencrouse, line 100:
```suggestion
```

PR #491, opened 2020-08-25
Add libvips for better image processing performance

bencrouse:
libvips offers dramatically improved performance using lower memory, so
this is the preferred image processing solution moving forward.

For upgrading, a configuration is provided to fallback to the old
ImageMagick implementation.


admin/test/integration/workarea/admin/jump_to_integration_test.rb
```diff
@@ -127,7 +127,7 @@ def test_finding_content_assets
 
         results = JSON.parse(response.body)['results']
         assert_equal(1, results.length)
-        assert_equal('Test Asset - file.jpg', results.first['label'])
+        assert_equal('Test Asset - product_image.jpg', results.first['label'])
```
tubbo, line 130: we fixed it lol

bencrouse: it was changed because the factory now has the filename.

PR #492, opened 2020-08-25
Shipping SKUs Admin UI

tubbo:
This adds an interface to Shipping SKUs from within the admin. Based on the Pricing SKUs admin UI, it allows for administration of a SKU's weight and dimensions. This information is used by ActiveShipping in order to generate a package and automatically obtain shipping rates for the item represented by this SKU.

### DIFF
--- a/admin/app/controllers/workarea/admin/shipping_skus_controller.rb
+++ b/admin/app/controllers/workarea/admin/shipping_skus_controller.rb
@@ -1,0 +1,53 @@
+module Workarea
+  module Admin
+    class ShippingSkusController < Admin::ApplicationController
+      required_permissions :catalog
+
+      after_action :track_index_filters, only: :index
+
+      def index
+        search = Search::AdminShippingSkus.new(view_model_options)
+        @search = SearchViewModel.new(search, view_model_options)
+      end
+
+      def new
+        @sku = ShippingSkuViewModel.wrap(Shipping::Sku.new)
+      end
+
+      def show
+        @sku = ShippingSkuViewModel.wrap(
+          Shipping::Sku.find_or_create_by(id: params[:id])
+        )
+      end
+
+      def create
+        @sku = Shipping::Sku.new(params[:sku])
+
+        if @sku.save
+          flash[:success] =
+            t('workarea.admin.shipping_skus.saved', sku: @sku.id)
+          redirect_to shipping_sku_path(@sku)
+        else
+          render :new, status: :unprocessable_entity
+        end
+      end
+
+      def edit
+        @sku = ShippingSkuViewModel.wrap(Shipping::Sku.find(params[:id]))
+      end
+
+      def update
+        @sku = Shipping::Sku.find(params[:id])
+
+        if @sku.update_attributes(params[:sku])
+          flash[:success] =
+            t('workarea.admin.shipping_skus.saved', sku: @sku.id)
+          redirect_to shipping_sku_path(@sku)
+        else
+          @sku = ShippingSkuViewModel.new(@sku)
+          render :edit, status: :unprocessable_entity
+        end
+      end
+    end
+  end
+end
--- a/admin/app/view_models/workarea/admin/shipping_sku_view_model.rb
+++ b/admin/app/view_models/workarea/admin/shipping_sku_view_model.rb
@@ -1,0 +1,34 @@
+module Workarea
+  module Admin
+    class ShippingSkuViewModel < ApplicationViewModel
+      def weight
+        I18n.t("workarea.admin.shipping_skus.unit.#{weight_units}", count: model.weight || 0)
+      end
+
+      def height
+        I18n.t("workarea.admin.shipping_skus.unit.#{length_units}", count: model.height || 0)
+      end
+
+      def width
+        I18n.t("workarea.admin.shipping_skus.unit.#{length_units}", count: model.width || 0)
+      end
+
+      def length
+        I18n.t("workarea.admin.shipping_skus.unit.#{length_units}", count: model.length || 0)
+      end
+
+      def timeline
+        @timeline ||= TimelineViewModel.new(model)
+      end
+
+      def product
+        return @product if defined?(@product)
+
+        @product = begin
+          product = Catalog::Product.find_by_sku(model.id)
+          ProductViewModel.wrap(product, options) if product.present?
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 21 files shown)


admin/app/views/workarea/admin/catalog_variants/index.html.haml
```diff
@@ -34,6 +34,7 @@
             %th= t('workarea.admin.fields.fulfillment')
             %th.align-center= t('workarea.admin.fields.inventory')
             %th.align-right= t('workarea.admin.fields.pricing')
+            %th.align-right= t('workarea.admin.fields.shipping')
```
tubbo, line 37: Right now, this just says "Shipping" to match the other column names like "Fulfillment", "Inventory", etc. But when you read it on the table, it doesn't really make much sense. In my opinion, a more descriptive name for this would be "Packaging", because the attributes expressed in a `Shipping::Sku` describe the weight and dimensions of a given SKU.

admin/app/views/workarea/admin/catalog_variants/index.html.haml
```diff
@@ -49,6 +50,11 @@
               %td.align-right
                 = link_to pricing_sku_path(variant.sku) do
                   = number_to_currency variant.pricing.sell_price
+              %td.align-right
+                = link_to shipping_sku_path(variant.sku), class: 'index__table-multi-line-link' do
+                  = variant.weight
+                  %br
+                  = variant.dimensions
```
mttdffy, line 57: Why does this need to be two lines? Not sure we need to present both pieces of data on this table, and if so maybe it should just be two columns `Weight` and `Dimensions` with each value linking to shipping sku.

admin/app/controllers/workarea/admin/shipping_skus_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+        else
+          @sku = ShippingSkuViewModel.new(@sku)
+          render :edit, status: :unprocessable_entity
+        end
+      end
+
+      private
+
+      def sku_params
+        params.require(:sku).permit(:id, :weight, :height, :width, :length)
```
bencrouse, line 55: The admin doesn't use strong parameters.

admin/app/views/workarea/admin/shipping_skus/edit.html.haml
```diff
@@ -0,0 +1,59 @@
+- @page_title = t('workarea.admin.shipping_skus.edit.title', name: @sku.name)
+
+.view
+  .view__header
+    .grid
+      .grid__cell.grid__cell--25
+        / left side element needed for spacing
```
tubbo, line 7: I know this looks a bit odd, but I really couldn't think of a better approach without diving too much into the CSS of `.grid`. Maybe something we can revisit later, but adding this spacer element is what prevents this:

Screen Shot 2020-08-27 at 9 18 07 AM

from looking like this:

Screen Shot 2020-08-27 at 9 17 57 AM

That looks like a bug on the inventory SKUs UI, but maybe I can address that in a later ticket...

admin/app/views/workarea/admin/shipping_skus/index.html.haml
```diff
@@ -0,0 +1,81 @@
+- @page_title = t('workarea.admin.shipping_skus.index.title')
+
+.view
+  .view__header
+    .grid.grid--middle.grid--center
```
tubbo, line 5: note: didn't need that spacer boi here because there was nothing to align left/right, just center.

admin/app/view_models/workarea/admin/shipping_sku_view_model.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      def timeline
+        @timeline ||= TimelineViewModel.new(model)
+      end
+
+      def product
+        @product ||=
+          begin
+            product = Catalog::Product.find_by_sku(model.id)
+            ProductViewModel.wrap(product, options) if product.present?
```
bencrouse, line 28: This memoizing won't work if this `begin` block returns `nil`. Go for the `if defined?` style.

admin/app/views/workarea/admin/shipping_skus/edit.html.haml
bencrouse: I believe you can fix this by adding the `grid--right` modifier to the grid instead of an extra `grid__cell`.

admin/app/views/workarea/admin/shipping_skus/show.html.haml
```diff
@@ -0,0 +1,16 @@
+- @page_title = @sku.name
+
+.view
+  .view__header
+    .grid
+      .grid__cell.grid__cell--25
+        / left side element needed for spacing
```
bencrouse, line 7: Here too

admin/config/locales/en.yml
```diff
@@ -3849,6 +3855,37 @@ en:
...
+        index:
+          button: Add New Shipping SKU
+          title: Shipping
+          sku_pluralize:
+            one: "%{count} SKU"
+            other: "%{count} SKUs"
+        new:
+          create_sku: Create Shipping SKU
+          title: New Shipping SKU
+        saved: "%{sku} has been saved."
```
bencrouse, line 3883: Conventionally, we put these in a `flash_messages` section.

admin/config/routes.rb
```diff
@@ -430,5 +430,7 @@
 
   resource :configuration, only: [:show, :update]
 
+  resources :shipping_skus, except: [:destroy]
```
bencrouse, line 433: Why can't these be destroyed?

core/app/models/workarea/shipping/sku.rb
```diff
@@ -11,18 +11,38 @@ class Sku
...
+      def name
+        I18n.t('workarea.shipping_sku.name', id: id)
+      end
+
       def height
         dimensions.first
       end
 
+      def height=(new_height)
+        self.dimensions[0] = new_height&.to_f
```
bencrouse, line 27: (minor) you don't need the `self` here.

admin/config/routes.rb
tubbo: they're destroyed through the bulk editing feature. there's also no `#destroy` action on the pricing skus controller.

admin/config/locales/en.yml
tubbo: i just copied this from the pricing skus controller, can change it if you want though. https://github.com/workarea-commerce/workarea/blob/master/admin/app/controllers/workarea/admin/pricing_skus_controller.rb#L43-L44

admin/app/views/workarea/admin/shipping_skus/show.html.haml
tubbo: ah perfect, thanks! I think I had `.grid--auto` or something and that was messing everything up.

PR #493, opened 2020-08-26
Fix Admin Configuration for Email Addresses

tubbo:
The hard-coded `config.email_from` and `config.email_to` settings conflict with the out-of-box administrable configuration for the "Email From" and "Email To" settings. This causes a warning for admins that explain why the settings on "Email To" and "Email From" won't take effect. Since the whole purpose of moving these settings to admin configuration was to let admins actually change them, the `config.email_from` and `config.email_to` settings have been removed from both default configuration and the `workarea:install` generator.

### DIFF
--- a/core/config/initializers/00_configuration.rb
+++ b/core/config/initializers/00_configuration.rb
@@ -156,14 +156,15 @@ Workarea::Configuration.define_fields do
   end
 
   fieldset 'Communication', namespaced: false do
+
     field 'Email From',
       type: :string,
-      default: 'noreply@example.com',
+      default: -> { "#{Workarea.config.site_name} <noreply@#{Workarea.config.host}>" },
       description: 'The email address used as the sender of system emails'
 
     field 'Email To',
       type: :string,
-      default: 'customerservice@example.com',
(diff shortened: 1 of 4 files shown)


tubbo:
image

core/config/initializers/00_configuration.rb
```diff
@@ -156,14 +156,15 @@
   end
 
   fieldset 'Communication', namespaced: false do
+
     field 'Email From',
       type: :string,
-      default: 'noreply@example.com',
+      default: -> { "#{Workarea.config.site_name} <noreply@#{Workarea.config.host}>" },
```
bencrouse, line 162: Is using a lambda actually supported? How did you test this?

mttdffy: default is passed through directly to the mongoid field definition, so if its supported by mongoid its supported here.

tubbo: Sure is! I tested it locally on my dev machine, first by restarting my server and then destroying the existing admin config that I had. When I went to the admin configuration page, I saw "noreply@store.test" as my Email From.

PR #495, opened 2020-08-27
Set Default Inventory Policy to "Standard" in Create Product Workflow

tubbo:
When creating a new product through the workflow, setting the "Inventory" on a particular SKU would still cause the `Inventory::Sku` to be created with the "Ignore" policy rather than "Standard". Setting inventory on a SKU now automatically causes the `Inventory::Sku` record to be created with a policy of "Standard" so as to deduct the given inventory to the varaint. When no inventory is given, Workarea will fall back to the default of "Ignore".

### DIFF
--- a/admin/app/controllers/workarea/admin/bulk_variant_saving.rb
+++ b/admin/app/controllers/workarea/admin/bulk_variant_saving.rb
@@ -21,8 +21,9 @@ module Workarea
         end
 
         if attributes[:inventory].present?
-          inventory = Inventory::Sku.find_or_create_by(id: variant.sku)
+          inventory = Inventory::Sku.find_or_initialize_by(id: variant.sku)
           inventory.available = attributes[:inventory]
+          inventory.policy = 'standard' if inventory.new_record?
           inventory.save!
         end
       end


admin/app/controllers/workarea/admin/bulk_variant_saving.rb
```diff
@@ -23,6 +23,7 @@ def save_variant_on_product(product, attributes: {}, variant: nil)
         if attributes[:inventory].present?
           inventory = Inventory::Sku.find_or_create_by(id: variant.sku)
           inventory.available = attributes[:inventory]
+          inventory.policy = 'standard'
```
bencrouse, line 26: So it looks like this is used in sequential editing as well, where the `Inventory::Sku` might already exist and so this would override whatever it was already set to. Can you put a check on this to only do this if the inventory is new?

Side note, is there a reason it uses `find_or_create_by` instead of `find_or_initialize_by`? That's just doing an extra db write.

tubbo: No idea, but I was thinking the same thing. Mind if I change that to `.find_or_initialize_by`?

PR #496, opened 2020-08-27
Ensure a supported version of libvips

bencrouse:
Trying to load the ruby-vips gem on an old version of libvips will
error, this shows a nicer warning.

### DIFF
--- a/core/lib/workarea/warnings.rb
+++ b/core/lib/workarea/warnings.rb
@@ -7,6 +7,7 @@ module Workarea
       check_mongo_notable_scan
       check_dragonfly_config
       check_auto_capture
+      check_libvips_version
     end
 
     def check_timezone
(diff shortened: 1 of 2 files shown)


core/lib/workarea/warnings.rb
```diff
@@ -86,6 +87,22 @@ def check_auto_capture
...
+    def check_libvips_version
+      if Configuration::ImageProcessing.libvips_version.present? &&
+          !Configuration::ImageProcessing.libvips?
+
+        warn <<~eos
+**************************************************
+⛔️ WARNING: libvips is installed but will not be used.
+
+The version of libvips installed is out of date. Workarea will fallback to using
+ImageMagick. We highly recommend upgrading to the latest verison of libvips to
```
tubbo, line 104:
```suggestion
ImageMagick. We highly recommend upgrading to the latest version of libvips to
```

tubbo, line 104:
```suggestion
ImageMagick. We highly recommend upgrading to the latest version of libvips to
```

core/lib/workarea/warnings.rb
```diff
@@ -86,6 +87,22 @@ def check_auto_capture
...
+      end
+    end
+
+    def check_libvips_version
+      if Configuration::ImageProcessing.libvips_version.present? &&
+          !Configuration::ImageProcessing.libvips?
+
+        warn <<~eos
+**************************************************
+⛔️ WARNING: libvips is installed but will not be used.
```
tubbo, line 101: Can we print the version here? Also, I think this would happen if you for some reason had libvips installed but `vips` wasn't in your `$PATH`. Might want to mention that's the case before people spend an hour recompiling something they didn't need to...

```suggestion
⛔️ WARNING: libvips (#{Configuration::ImageProcessing.libvips_version}) is installed in your $PATH but will not be used.
```

bencrouse, line 101:
```suggestion
⛔️ WARNING: #{Configuration::ImageProcessing.libvips_version} is available but will not be used.
```

PR #498, opened 2020-08-31
Update admin views for consistent display of inventory availability

mttdffy:
WORKAREA-262

### DIFF
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -102,7 +102,7 @@ en:
       name: "Fulfillment SKU %{id}"
     inventory_sku:
       name: "Inventory %{id}"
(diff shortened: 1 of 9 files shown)


mttdffy:
This could be a patch, but for now I have it pointed at master for 3.6. Thoughts?

core/config/locales/en.yml
```diff
@@ -102,6 +102,7 @@ en:
       name: "Fulfillment SKU %{id}"
     inventory_sku:
       name: "Inventory %{id}"
+      jump_to_text: "%{id} (%{count} sellable units)"
```
bencrouse, line 105:
```suggestion
      jump_to_text: "%{id} (%{count} sellable)"
```

PR #500, opened 2020-09-01
Make Workarea compatible with Zeitwerk

bencrouse:
This updates to the latest version of the rails-decorators gem, and
makes a few minor changes to allow Zeitwerk loading to function
correctly.

To check that your project is good to go for Zeitwerk, we recommend
running `rails zeitwerk:check` which will test your application to
determine if it's ready.


core/app/models/workarea/search/storefront/product.rb
```diff
@@ -1,7 +1,7 @@
 module Workarea
   module Search
     class Storefront
-      class Product < Search::Storefront
+      class Product < ::Workarea::Search::Storefront
```
tubbo, line 4: We're using this same abbreviated syntax in all of the `search/admin` models, I wonder why you didn't need to do this same change in those classes?

bencrouse: Maybe something to do with the name `Product`?

PR #503, opened 2020-09-03
Fix incorrect tracking and metrics after impersonation

bencrouse:
Not managing the email cookie and unintentional merging of metrics leads
to incorrect values in the admin.


mttdffy (approved):

mttdffy (changes requested):
Relevant test failure?

tubbo (approved):

PR #504, opened 2020-09-03
Add config to allow defining a default tax code for shipping services

mttdffy:
WORKAREA-256

### DIFF
--- a/core/app/models/workarea/shipping/service.rb
+++ b/core/app/models/workarea/shipping/service.rb
@@ -37,15 +37,22 @@ module Workarea
       end
 
       def self.by_price(price)
-        cache.select do |method|
-          (method.subtotal_min.nil? || method.subtotal_min <= price) &&
-            (method.subtotal_max.nil? || method.subtotal_max >= price)
+        cache.select do |service|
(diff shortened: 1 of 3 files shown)


core/lib/workarea/configuration.rb
```diff
@@ -1306,6 +1306,12 @@ def self.setup_defaults
         cvv
         amount
       ]
+
+      # Tax code assigned to shipping options when an existing service does
+      # not exist. This is useful for third-party gateways to assign tax codes
+      # to dynamically generated options. This can either be a string or a proc.
+      # If using a proc, the carrier and name of the service will be passed.
+      config.default_shipping_service_tax_code = nil
```
bencrouse, line 1314: Any reason we shouldn't make this admin facing?

mttdffy: admin can't define procs?

bencrouse: the code looks like it would support both!

mttdffy: Not sure what you're saying

bencrouse (approved):

PR #507, opened 2020-09-09
Improve display of disabled toggles

bencrouse:
When a toggle button is disabled, it should reflect that visually
instead of just looking like it should be functional.


tubbo (approved):

bencrouse:
Updated for v3.4

PR #508, opened 2020-09-10
Set the default option of allow blank to true

bencrouse:
This is more consistent with the way models work, and has already been
assumed by a couple devs.

### DIFF
--- a/core/app/models/workarea/configuration/params.rb
+++ b/core/app/models/workarea/configuration/params.rb
@@ -17,12 +17,7 @@ module Workarea
               value
             end
 
-          memo[field.key] =
-            if formatted_value.blank? && !field.allow_blank?
-              field.default
-            else
-              formatted_value
-            end
+          memo[field.key] = formatted_value.presence || field.default
         end
       end
 
--- a/core/lib/workarea/configuration/administrable/field.rb
+++ b/core/lib/workarea/configuration/administrable/field.rb
@@ -53,7 +53,7 @@ module Workarea
         end
 
         def allow_blank?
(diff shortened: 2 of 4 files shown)


tubbo (approved):

core/app/models/workarea/configuration/params.rb
```diff
@@ -17,12 +17,7 @@ def to_h
               value
             end
 
-          memo[field.key] =
-            if formatted_value.blank? && !field.allow_blank?
-              field.default
-            else
-              formatted_value
-            end
+          memo[field.key] = formatted_value.presence || field.default
```
mttdffy, line 20: so if someone defines a field

```
field 'Foo',
  type: String,
  default: 'bar',
  allow_blank: true
```
and the admin blanks out the field, the value will be `foo`?

mttdffy (changes requested):
Think this is potentially dangerous to flip this in a patch

core/app/models/workarea/configuration/params.rb
bencrouse: Isn't that what a default means?

mttdffy: default is the initial value, a user blanking out a field that is allowed to be blank seems intentional enough that it should be left blank. If the field required a value than setting default when left blank would make sense but not in the same of a field that the developer defined as ok to be blank.

bencrouse:
If we don't want to change behavior in a patch, this really can't be achieved. I'll open different PRs.

PR #511, opened 2020-09-11
Add content security policy controls

mttdffy:
WORKAREA-268

### DIFF
--- a/admin/config/locales/en.yml
+++ b/admin/config/locales/en.yml
@@ -672,6 +672,7 @@ en:
           browser_title: Browser Title
           browser_title_note: The title appears in the top bar of a web browser and is shown in search engine results.
           browser_title_placeholder: "(defaults to content name)"
+          content_security_policy_note: This should be a single line string representing the content security policy for this content. This will be appended to the global security policy, if any.
           clear_asset: Clear asset
           css: CSS
           css_note: The CSS block is output just before the end of the head. The content of this block is automatically is wrapped in a style tags.
@@ -1644,6 +1645,7 @@ en:
         backordered_units: Backordered Units
         backordered_until: Backordered Until
         body: Body
+        blocked_uri: Blocked URI
         breadcrumbs: Breadcrumbs
         browser_info: Browser Info
         browser_title: Browser Title
@@ -1661,6 +1663,7 @@ en:
         completed_at: Completed
         content: Content
         content_blocks: Content Blocks
+        content_security_policy: Content Security Policy
         copied_from_id: Copied From
         count: Count
         country: Country
@@ -1684,6 +1687,7 @@ en:
         discount: Discount
         discounts: Discounts
         display_name: Display Name
+        document_uri: Document URI
         downloads: Downloads
         email: Email
         enabled: Enabled
@@ -1747,6 +1751,7 @@ en:
         order: Order
         orders: Orders
         original_id: Original ID
+        original_policy: Policy
         other: "%{count} fields"
         packaged_product_ids: Packaged Products
         pass: Pass
@@ -1856,6 +1861,7 @@ en:
         url: URL
         user_tags: User Tags
         variants: Variants
+        violated_directive: Violated Directive
         weight: Weight
         width: Width
       form:
@@ -3278,6 +3284,8 @@ en:
           title: Average Order Value
         back: Back
         by: by
+        content_security_policy_violations:
+          title: Content Security Policy Violations
         customers:
           filters:
             all: All Customers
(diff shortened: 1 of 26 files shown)


admin/config/locales/en.yml
```diff
@@ -672,6 +672,7 @@ en:
           browser_title: Browser Title
           browser_title_note: The title appears in the top bar of a web browser and is shown in search engine results.
           browser_title_placeholder: "(defaults to content name)"
+          content_security_policy_note: This should be a single line string representing the content security policy for this content. This will be appended to the global security policy, if any.
```
bencrouse, line 675: Can you add an example here for them to reference? Maybe we should move to a tooltip to allow more thorough explanation and like a copy pasta example.

core/config/initializers/13_rack_attack.rb
```diff
@@ -182,6 +182,25 @@ class Rack::Attack
     end
   end
 
+  # Throttle POST requests to /content_security_violations by IP address
+  #
+  # Key: "rack::attack:#{Time.now.to_i/:period}:csp_violations/ip:#{req.ip}"
+  throttle('csp_violations/ip', limit: 10, period: 10.minutes) do |request|
```
bencrouse, line 188: Seems good, I'm curious about the thought to throttle this harder than the general throttle.

storefront/.byebug_history
```diff
@@ -0,0 +1,12 @@
+continue
```
bencrouse, line 1: Debugging!

storefront/app/controllers/workarea/storefront/content_security_policy.rb
```diff
@@ -0,0 +1,52 @@
+module Workarea
+  module Storefront
+    module ContentSecurityPolicy
```
bencrouse, line 3: It looks like the Rails DSL supports a block that can be dynamic in a controller. Can we use that instead of the artisanal approach?

storefront/app/controllers/workarea/storefront/categories_controller.rb
```diff
@@ -7,5 +7,11 @@ def show
       raise InvalidDisplay unless model.active? || current_user.try(:admin?)
       @category = Storefront::CategoryViewModel.new(model, view_model_options)
     end
+
+    private
+
+    def page_specific_content_policy
```
bencrouse, line 13: This would be nicer as a delegation, something like `delegate :content_security_policy_config, to: :@category`

storefront/app/controllers/workarea/storefront/pages_controller.rb
```diff
@@ -19,5 +19,11 @@ def robots; end
     def accessibility; end
     def browser_config; end
     def web_manifest; end
+
+    private
+
+    def page_specific_content_policy
```
bencrouse, line 25: Delegation would be less lines of code!

bencrouse (changes requested):
Also can you add a system test for the report?

storefront/app/controllers/workarea/storefront/content_security_policy.rb
mttdffy: but the block needs to use the dsl, which would be a nightmare to support via user input.

storefront/app/controllers/workarea/storefront/categories_controller.rb
mttdffy: I guess? Not 💯 on what you mean.

core/config/initializers/13_rack_attack.rb
mttdffy: In my head its likely that if violations were present on the site, a single user would hit the same violations over and over. So basically I set the limit to a number thats reasonable but stops the same violation reports from getting processed repeatedly.

storefront/app/controllers/workarea/storefront/pages_controller.rb
mttdffy: in the controller, but I'd have to define a method elsewhere. Don't think it's worth it.

storefront/app/controllers/workarea/storefront/categories_controller.rb
bencrouse: We can rename this method (and the one on the model) to `content_security_policy_config` or something similar and use the above delegation rather than define the "extra" method here.

storefront/app/controllers/workarea/storefront/content_security_policy.rb
bencrouse: Like because the admin input is just raw text?

bencrouse (changes requested):

bencrouse:
Maybe we should have a UX discussion over this?

bencrouse:
Closing, for now.

PR #513, opened 2020-09-14
Add required option to config DSL in place of allow blank

bencrouse:
This removes the `allow_blank` option from the configuration DSL and adds
the `required` in its place. This is done to align with developer
expectations we've experienced.

### DIFF
--- a/core/app/models/workarea/configuration/params.rb
+++ b/core/app/models/workarea/configuration/params.rb
@@ -17,12 +17,7 @@ module Workarea
               value
             end
 
-          memo[field.key] =
-            if formatted_value.blank? && !field.allow_blank?
-              field.default
-            else
-              formatted_value
-            end
+          memo[field.key] = formatted_value.presence || field.default
         end
       end
 
--- a/core/app/models/workarea/configuration/params.rb
+++ b/core/app/models/workarea/configuration/params.rb
@@ -17,7 +17,11 @@ module Workarea
               value
             end
 
-          memo[field.key] = formatted_value.presence || field.default
+          memo[field.key] = if field.required?
+            formatted_value.presence || field.default
+          else
+            formatted_value || field.default
+          end
         end
       end
 
--- a/docs/source/articles/configuration-fields.html.md
+++ b/docs/source/articles/configuration-fields.html.md
@@ -98,13 +98,13 @@ Of note, `duration` allows you to define a field for periods of time like `3.day
 
 When defining a configuration field, there are a number of options you can provide.
 
-- `allow_blank`
+- `required`
 
-  Defaults to false. Whether or not the field can be set to a blank value by a user. Blank values submitted for a field that does not allow it will be set to the field's default value. The admin UI will also prevent user's from submitting blank field values when this is false.
+  Whether or not the field is required, defaults to `false`. Blank values submitted for a field that does not allow it will be set to the field's default value. The admin UI will also prevent user's from submitting blank field values when this is `false`.
 
 - `default`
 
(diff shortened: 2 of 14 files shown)


tubbo (approved):

core/app/models/workarea/configuration/params.rb
```diff
@@ -17,12 +17,7 @@ def to_h
               value
             end
 
-          memo[field.key] =
-            if formatted_value.blank? && !field.allow_blank?
-              field.default
-            else
-              formatted_value
-            end
+          memo[field.key] = formatted_value.presence || field.default
```
mttdffy, line 20: Did we ever come to resolution with this from the old PR? Still think this logic is no good in the scenario of field that is not required, with a default value, but an admin wants to it be blank.

docs/source/articles/configuration-fields.html.md
```diff
@@ -98,13 +98,13 @@ Of note, `duration` allows you to define a field for periods of time like `3.day
 
 When defining a configuration field, there are a number of options you can provide.
 
-- `allow_blank`
+- `required`
 
-  Defaults to false. Whether or not the field can be set to a blank value by a user. Blank values submitted for a field that does not allow it will be set to the field's default value. The admin UI will also prevent user's from submitting blank field values when this is false.
+  Whether or not the field is required, defaults to `false`. Blank values submitted for a field that does not allow it will be set to the field's default value. The admin UI will also prevent user's from submitting blank field values when this is `false`.
```
mttdffy, line 103: This set of sentences is really confusing in the context of `requires` over the old `allow_blank`

```suggestion
  Whether or not the field is required, defaults to `false`. The admin UI will prevent user's from submitting blank field values for required fields. If values are submitted in some other way, blank values for a required field will be set to the field's default value.
```

docs/source/articles/configuration-fields.html.md
```diff
@@ -98,13 +98,13 @@ Of note, `duration` allows you to define a field for periods of time like `3.day
...
 
 - `default`
 
-  The initial value of the field when newly created. Must match the type defined for the field.
+  The value that will be returned when there isn't a specified value or it is blank. Must match the type defined for the field.
```
mttdffy, line 107: No likey! I don't think an admin setting a non-required field blank is unreasonable.

mttdffy (changes requested):
I think if we are making required `false` by default, we need to update the existing fields defined in base as required, yes?

bencrouse:
Regarding the current field definitions, all the ones that seem required have defaults, so there wasn't a need to add `required`.

core/app/models/workarea/configuration/params.rb
bencrouse: Oh that must have happened while you were out.

https://github.com/workarea-commerce/workarea/pull/508
https://github.com/workarea-commerce/workarea/pull/512

Logically, there wasn't a way to change the default of `allow_blank` and preserve all the existing functionality, which you felt shouldn't be changed in a patch.

docs/source/articles/configuration-fields.html.md
bencrouse: I'm not so sure, but if you did want to achieve that, you could just not provide a `default` and set `required: false`.

bencrouse:
I think there's kind of 2 ways to think about what this feature is and that's leading to the confusion. One is that this is a model-like DSL to configure the `Configuration::Admin` instance. The other is a DSL to configure the admin UI. Maybe reconciling that will help us find clarity in the way we think about it.

mttdffy (approved):

PR #514, opened 2020-09-14
Redirect back to the previous page after stopping impersonation

bencrouse:
Currently we redirect to the user's show page, which can be pretty
disorienting.

### DIFF
--- a/admin/config/locales/en.yml
+++ b/admin/config/locales/en.yml
@@ -3512,7 +3512,7 @@ en:
           changes_saved: Your changes have been saved
           saved: Your changes have been saved
           started: You are now browsing as %{email}
-          stopped: Impersonation has been stopped, you're browsing as your account.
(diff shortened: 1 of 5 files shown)


admin/config/locales/en.yml
```diff
@@ -3512,7 +3512,7 @@ en:
           changes_saved: Your changes have been saved
           saved: Your changes have been saved
           started: You are now browsing as %{email}
-          stopped: Impersonation for this user has been stopped.
+          stopped: Impersonation has been stopped, you're browsing as your account.
```
tubbo, line 3515:
```suggestion
          stopped: Impersonation has been stopped, you are now browsing as your account.
```

tubbo (changes requested):

mttdffy (approved):

PR #515, opened 2020-09-17
Change admin config field DSL to define fields by ID rather than name

mttdffy:
WORKAREA-275

I'll have to update this with the PR for require, but its done so I opened this.


bencrouse:
Looks good to me.

bencrouse:
So I merged #513, no conflicts here, was there anything else you wanted to check before merging?

mttdffy:
> So I merged #513, no conflicts here, was there anything else you wanted to check before merging?

No. I thought for sure there would be issues. I think I assumed you'd be adding `required: true` to all the existing fields in `00_configuration.rb`. We should do that, no?

bencrouse:
Yeah, I'll update.

bencrouse (approved):

PR #516, opened 2020-09-17
Add sitemap generation

mttdffy:
WORKAREA-255

### DIFF
--- a/core/app/models/workarea/navigation/taxon.rb
+++ b/core/app/models/workarea/navigation/taxon.rb
@@ -126,6 +126,10 @@ module Workarea
         navigable.is_a?(SearchResults)
       end
 
+      def show_in_sitemap?
+        !placeholder? && active? && (navigable.present? || url.to_s.start_with?('http', '/'))
+      end
+
       private
 
       def set_parent
--- a/core/app/models/workarea/sitemap/link.rb
+++ b/core/app/models/workarea/sitemap/link.rb
@@ -1,0 +1,53 @@
+module Workarea
+  class Sitemap
+    # Model object for representing a single link in the sitemap. A
+    # sitemap XML file is composed of +<url />+ elements, and this class
+    # helps to define the data that ends up in there.
+    class Link
+      delegate_missing_to :@taxon
+
+      # @param [Workarea::Navigation::Taxon] taxon - Model this link is based on
+      # @param [Workarea::GenerateSitemaps] generator - Worker class this link was instantiated from.
+      def initialize(taxon:, generator:)
+        @taxon = taxon
+        @generator = generator
+      end
+
+      # Return the parsed +host+ from the given URL, or the default host
+      # (+Workarea.config.host+) if a +Navigable+ model is being linked
+      # to.
+      def host
+        return default_host unless fully_qualified_url?
+        "#{uri.scheme}://#{uri.host}"
+      end
+
+      # Return the route path for the +Navigable+ model that is linked
+      # to in this taxon, or the +url+ that was hard-coded into the
+      # taxon at creation.
+      def path
+        return uri.path if url?
+
+        @generator.send(route, navigable_slug)
+      end
+
+      private
+
+      def route
+        @route ||= "#{resource_name}_path"
+      end
+
+      def uri
+        @uri ||= URI.parse(url)
+      end
+
+      def default_host
+        scheme = Rails.configuration.force_ssl ? 'https' : 'http'
+        "#{scheme}://#{Workarea.config.host}"
+      end
+
+      def fully_qualified_url?
+        url? && uri.host.present? && uri.scheme.present?
+      end
+    end
+  end
+end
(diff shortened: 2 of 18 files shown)


core/app/models/workarea/navigation/taxon.rb
```diff
@@ -126,6 +126,10 @@ def search_results?
         navigable.is_a?(SearchResults)
       end
 
+      def show_in_sitemap?
```
bencrouse, line 129: So now that we have this method, it feels like the HTML sitemap should also respect this as well as the XML one. Can you update that?

core/app/workers/workarea/generate_sitemaps.rb
```diff
@@ -0,0 +1,71 @@
+module Workarea
+  class GenerateSitemaps
+    include Sidekiq::Worker
+    include Storefront::Engine.routes.url_helpers
+
+    TMP_DIR = Rails.root.join('tmp', 'sitemaps')
```
bencrouse, line 6: It'd be safer (for handling concurrency and whatnot) to not have this be a constant (unique per instance), and to use a built in Ruby temp dir which can give you a unique dir.

core/config/initializers/27_sitemaps.rb
```diff
@@ -0,0 +1,3 @@
+# Quiet the output of the generator in tests to eliminate
+# the noise when running the test suite.
+SitemapGenerator.verbose = !Rails.env.test?
```
bencrouse, line 3: YAS

core/config/initializers/07_dragonfly.rb
```diff
@@ -17,6 +17,14 @@
 
   url_format '/media/:job/:name'
   processor :optim, Workarea::ImageOptimProcessor
+
+  response_header 'Cache-Control' do |job, request, headers|
+    if request.path =~ /sitemap/
+      'public, max-age=86400' # 1 day (equal to sitemap generation frequency)
+    else
+      'public, max-age=31536000' # 1 year (Dragonfly default)
```
bencrouse, line 25: Maybe these times should be configs? Thoughts?

storefront/app/views/workarea/storefront/pages/robots.text.erb
```diff
@@ -1,5 +1,6 @@
 User-Agent: *
 Allow: <%= sitemap_path %>
+Sitemap: https://<%= Workarea.config.host %>/sitemap.xml.gz
```
bencrouse, line 3: We really need a `Workarea.protocol`

bencrouse (changes requested):

core/config/initializers/27_sitemaps.rb
mttdffy: not sure why we didn't do this before

core/app/workers/workarea/generate_sitemaps.rb
mttdffy: Agreed. Honestly didn't even look at the code very hard.

core/config/initializers/07_dragonfly.rb
mttdffy: i'd lean towards no because the sitemap caching is set to mimic the frequency in which it might change, and the other is a default no one has ever asked or seemed to care about.

bencrouse: Yeah fair enough.

bencrouse (changes requested):
Looks good, one more thought: can you add a warning that gets output if you have the sitemaps plugin installed? I'm thinking about the upgrade path, we wouldn't want people to still have that installed. Maybe a note for the upgrade guide would be good too.

core/app/models/workarea/sitemap/link.rb
```diff
@@ -0,0 +1,53 @@
+module Workarea
+  class Sitemap
+    # Model object for representing a single link in the sitemap. A
+    # sitemap XML file is composed of +<url />+ elements, and this class
+    # helps to define the data that ends up in there.
+    class Link
```
tubbo, line 6: hey i remember this class :)

core/config/initializers/27_sitemaps.rb
tubbo: i literally had no idea this existed otherwise i definitely would have!

storefront/app/views/workarea/storefront/pages/robots.text.erb
tubbo: that would be really nice. when i was working with site-builder on `https://mysitebuildersite.test`, my previews would look terrible because the CSS/JS assets were trying to be fetched from `http://somesite-preview.mysitebuildersite.test`. developers using puma-dev are going to hit that in the real world if they use an https URL...

tubbo (approved):

bencrouse (approved):

bencrouse (changes requested):
I missed the build failures, they look relevant to these changes.

bencrouse (approved):

PR #517, opened 2020-09-22
Fix Precision of Tax Rates UI

tubbo:
The `:step` values of the new/edit forms and precision configuration for `#number_to_percentage` were not only rounding certain tax rates to an incorrect number, but were also showing a bunch of insignificant zeroes in the admin for tax rates. To resolve this, configure `#number_to_percentage` to have 3-decimal precision, and strip all insignificant zeroes from the display, leaving the admin with a much nicer percentage display than what was presented before.

**Before:**

before

**After:**

after

Here's another pic of all the different precision levels:

Screen Shot 2020-09-22 at 10 15 45 AM

### DIFF
--- a/admin/app/views/workarea/admin/tax_rates/edit.html.haml
+++ b/admin/app/views/workarea/admin/tax_rates/edit.html.haml
@@ -21,7 +21,7 @@
         .grid__cell.grid__cell--50
           .property
             = label_tag 'rate_country_percentage', t('workarea.admin.tax_rates.index.table.country_percentage'), class: 'property__name'
-            = number_field_tag 'rate[country_percentage]', @rate.country_percentage * 100, min: 0, max: 100, step: 0.01, class: 'text-box text-box--i18n'
+            = number_field_tag 'rate[country_percentage]', @rate.country_percentage * 100, min: 0, max: 100, step: '0.001', class: 'text-box text-box--i18n'
 
         .grid__cell.grid__cell--50
           .property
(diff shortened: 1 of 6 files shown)


admin/app/views/workarea/admin/tax_rates/edit.html.haml
```diff
@@ -21,7 +21,7 @@
         .grid__cell.grid__cell--50
           .property
             = label_tag 'rate_country_percentage', t('workarea.admin.tax_rates.index.table.country_percentage'), class: 'property__name'
-            = number_field_tag 'rate[country_percentage]', @rate.country_percentage * 100, min: 0, max: 100, step: 0.01, class: 'text-box text-box--i18n'
+            = number_field_tag 'rate[country_percentage]', @rate.country_percentage * 100, min: 0, max: 100, step: 'any', class: 'text-box text-box--i18n'
```
tubbo, line 24: The default for these fields is `1`, so making this "any" allows any kind of numerical input. This should resolve any current and future issues surrounding editing taxes in admin...

admin/app/views/workarea/admin/tax_rates/index.html.haml
```diff
@@ -56,9 +56,9 @@
               %td.align-center= rate.country.try(&:alpha2) || '-'
               %td.align-center= rate.region || '-'
               %td.align-center= rate.postal_code.presence || '-'
-              %td.align-center= number_to_percentage(rate.country_percentage * 100, precision: 1)
-              %td.align-center= number_to_percentage(rate.region_percentage * 100, precision: 1)
-              %td.align-center= number_to_percentage(rate.postal_code_percentage * 100, precision: 1)
+              %td.align-center= number_to_percentage(rate.country_percentage * 100, precision: 3, strip_insignificant_zeros: true)
```
tubbo, line 59: This is actually the default precision, but perhaps this should be configurable? It's the only real limiting factor that would potentially block you from seeing the exact input you entered in the form. Entering in a number like `7.0005` would result in the following admin UI:

Screen Shot 2020-09-22 at 10 21 24 AM

This is a bit strange to me, so I was thinking this could be an admin-configurable setting?

tubbo:
Card UI as well:

Screen Shot 2020-09-22 at 10 56 33 AM

bencrouse:
Any reason this is a draft?

bencrouse (approved):

PR #518, opened 2020-09-23
Improve clarity of discount verbiage

bencrouse:


admin/app/views/workarea/admin/pricing_discounts/properties/_product_attribute.html.haml
```diff
@@ -5,7 +5,7 @@
     %span.discount__node
       %span.property.property--inline
         = text_field_tag 'discount[amount]', @discount.amount, class: 'text-box text-box--small', title: 'Amount of Discount', placeholder: '10', required: true
-    %span.discount__node off products
+    %span.discount__node off the unit price
```
mttdffy, line 8: so it reads something like "take $20 off the unit price where brand is Nike..."?

Seems like it should be "take $20 off the unit price of products where brand is Nike..."

bencrouse: What else does "unit price" mean here?

tubbo: I think what's more unclear is what the condition of the discount is, rather than what we choose to call the price. Saying "$20 off the unit price where brand is Nike" is a bit ambiguous at first glance, because by itself it doesn't specify where the `brand: "Nike"` data point lives. Is that an item customization or a product filter? When you go full tunnel-vision and read this sentence without any other context, it's not very clear.

If the discount type is somewhere on the page, so the user knows they are editing a "Product Attribute" discount, then I think this is fine to leave as-is. I can't remember what this page looks like off the top of my head, but I feel like the actual type of the discount isn't really specified anywhere, so I do feel like making this line of text a bit more clear is worth it.

tubbo (changes requested):

admin/app/views/workarea/admin/pricing_discounts/properties/_product_attribute.html.haml
bencrouse: @mttdffy ?

mttdffy: Does the unit price have a brand of Nike? Yes its a safe implication you're talking about the product but I don't think the vagueness serves a purpose and adding "of products" only improves the readability.

mttdffy (approved):

tubbo (approved):

PR #520, opened 2020-09-23
Add Sequential Order Numbering Guide

tubbo:
Add a section to the developer docs explaining how Workarea constructs order IDs, why sequential order numbering is generally ill-advised, as well as a solution for implementing the "vanity" side of the feature.

### DIFF
--- a/docs/source/articles/sequential-order-numbering.html.md
+++ b/docs/source/articles/sequential-order-numbering.html.md
@@ -1,0 +1,81 @@
+---
+title: Sequential Order Numbering
+created_at: 2020/09/24
+excerpt: "In Workarea, orders (and other persisted data) are identified in a different manner from relational databases in order to avoid collisions. This guide explains how Workarea identifies orders, and how you might want to customize this for vanity purposes."
+---
+
+# Sequential Order Numbering
+
+In relational database systems, identifiers are typically numbers that are automatically incremented. Many ecommerce systems, both in the past and still running today, are built around relational databases. These databases stored IDs numerically, because that was the cheapest and most efficient way to do so, especially since at the time storage was not as cheap as it is today.
+
+After using Workarea for a bit, you may be wondering why it doesn't do things this way. We understand many merchants and vendors in the ecommerce world are used to seeing numerical, incrementing order numbers. However, Workarea uses MongoDB (a distributed database) to store its persistent data, and so identifying orders by an automatically incrementing number becomes more challenging to keep consistent. In Workarea, orders and other persisted data are identified in a different manner to avoid race conditions and colliding IDs as the data is being replicated to multiple machines.
+
+Some merchants may want orders having a numerical, incrementing identifier for their own vanity purposes. Many Workarea developers have found that explaining the rationale behind Workarea's order IDs has helped them get over this. This guide will go into detail about that rationale, explaining how orders are identified in the Workarea platform. For those merchants who insist on having this as a feature, this guide will also show developers how to implement sequential order numbering in a Workarea project, cleanly and safely. Although Workarea strongly discourages this practice, we believe the strategy described in this guide is the most reliable way of achieving the goal of sequential order numbering.
+
+## How Workarea Identifies Orders
+
+When an order is created or instantiated, it is assigned a randomized 5-digit uppercase hexadecimal value as its ID. Workarea does this so order numbers are legible to human eyes, and will play nicely on MongoDB. This ID is used throughout the system to link orders to their Payment and Shipping information, as well as in the admin for display purposes. Although this order ID is "randomized", `SecureRandom` generation uses the current system time to determine a seed value for its algorithm, so when compared, they are actually sequential already:
+
+```ruby
+def test_sequential_order_numbering
+  orders = [create_order, create_order]
+
+  assert(orders.second.id > orders.first.id) # => true
+end
+```
+
+However, Workarea's order IDs definitely aren't guaranteed to be numerical, and may look different from what merchants are used to.
+
+## Customizing Sequential Order Numbers
+
+Some clients may want this display ID to be sequentially numbered. This is possible to customize, but there's one caveat to keep in mind when doing so: MongoDB is a **distributed database**, so in production there are many instances of Mongo running to make sure that your app's data is not only within reach, but can be retrieved efficiently. This can lead to gaps in the order numbers, as well as order numbers appearing out of order in the admin (since Workarea only shows you orders which have been placed).
+
+If you do wish to customize this, however, we advise using the [Mongoid::Autoinc](https://github.com/suweller/mongoid-autoinc) plugin and decorating `Workarea::Order` like so:
+
+```ruby
+module Workarea
+  decorate Order do
+    decorated do
+      include Mongoid::Autoinc
+
+      field :number, type: Integer, default: 0
+
+      increments :number
+
+      # This will show the order number in the title of admin views.
+      def name
+        "#{super} (#{number})"
+      end
+    end
+  end
+end
+```
+
+You will probably also want to decorate places in the admin where this number would appear, such as the orders index page:
+
+```diff
+diff --git a/admin/app/views/workarea/admin/orders/index.html.haml b/admin/app/views/workarea/admin/orders/index.html.haml
+index c6c4cb7d4e..d00f9f67de 100644
+--- a/admin/app/views/workarea/admin/orders/index.html.haml
++++ b/admin/app/views/workarea/admin/orders/index.html.haml
+@@ -87,6 +87,7 @@
+                 = check_box_tag 'select_all', nil, false, id: 'select_all', class: 'checkbox__input', data: { bulk_action_select_all: '' }
+                 = label_tag 'select_all', t('workarea.admin.bulk_actions.select_all'), class: 'checkbox__label'
+             %th= t('workarea.admin.fields.id')
++            %th= t('workarea.admin.fields.number')
+             %th= t('workarea.admin.fields.email')
+             %th.align-right= t('workarea.admin.fields.total_price')
+             %th= t('workarea.admin.fields.payment_status')
+@@ -102,6 +103,7 @@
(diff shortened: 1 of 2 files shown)


docs/source/articles/orders-and-items.html.md
```diff
@@ -20,6 +20,55 @@ The `Order` model and its embedded items identify the consumer and the items tha
 - Zero or more promo codes, which may affect item and order pricing
 - The calculated order-level price totals
 
+## Order Identification
+
+When an order is created or instantiated, it is assigned a randomized 5-digit uppercase hexadecimal value as its ID. This is so order numbers are easy to read and identify, and will play nicely on a distributed database system such as MongoDB. This ID is used throughout the system to link orders to their Payment and Shipping information, as well as in the admin for display purposes. Some clients may want this display ID to be sequentially numbered, which is possible to customize, but we strongly advise against it. MongoDB is a distributed database, so in production there are many instances of Mongo running to make sure that your app's data is not only within reach, but can be retrieved efficiently. Because of this, **it is impossible to guarantee that two simultaneous writes for the same data will result in two synchronous operations within the database itself**.
```
jhill, line 25: Only the primary MongoDB replica set member receives writes. 

Tom, I don't think that you were in our IT discussions about this, but the only issue with the plan below is that there may be gaps in the sequence of order numbers, but not duplicates. All merchants that have had this implemented have been fine with some sporadic gaps.

bencrouse: It's also important to note that the sequence may or may not follow the chronology of when the orders were placed - order 10001 may be placed after order 10002.

bencrouse: Also, "easy to read and identify" is exactly why they want sequential numbers, not the random string ones.

bencrouse (changes requested):
Two things form our team discussions on this:
* Let's make it a separate article to call it out a bit more
* We discussed opening the article with a discussion about people's expectations coming from the relational database systems they're used to empathize with the people interested in this.

docs/source/articles/orders-and-items.html.md
tubbo: I guess I was trying to emphasize that our order numbers are the way they are _because_ they are easier to read and identify than a traditional BSON ID. Otherwise, what's the point? Why not just use a BSON ID?

I'll make the changes around sporadic gaps rather than simultaneous writes, and notate that the orders might not be placed in sequential order.

bencrouse (changes requested):
"We discussed opening the article with a discussion about people's expectations coming from the relational database systems they're used to empathize with the people interested in this."

Please open the article like we talked about.

docs/source/articles/sequential-order-numbering.html.md
```diff
@@ -0,0 +1,77 @@
...
+
+## How Workarea Identifies Orders
+
+When an order is created or instantiated, it is assigned a randomized 5-digit uppercase hexadecimal value as its ID. Workarea does this so order numbers are legible to human eyes, and will play nicely on a distributed database system such as MongoDB. This ID is used throughout the system to link orders to their Payment and Shipping information, as well as in the admin for display purposes. Although this order ID is "randomized", `SecureRandom` generation uses the current system time to determine a seed value for its algorithm, so when compared, they are actually sequential already:
```
bencrouse, line 13:
```suggestion
When an order is created or instantiated, it is assigned a randomized 5-digit uppercase hexadecimal value as its ID. Workarea does this so order numbers are legible to human eyes, and will play nicely on MongoDB. This ID is used throughout the system to link orders to their Payment and Shipping information, as well as in the admin for display purposes. Although this order ID is "randomized", `SecureRandom` generation uses the current system time to determine a seed value for its algorithm, so when compared, they are actually sequential already:
```

Brevity!

docs/source/articles/sequential-order-numbering.html.md
```diff
@@ -0,0 +1,77 @@
...
+
+```ruby
+def test_sequential_order_numbering
+  orders = [create_order, create_order]
+
+  assert(orders.second < orders.first) # => true
```
bencrouse, line 19: Shouldn't this be:

```suggestion
  assert(orders.second.id < orders.first.id) # => true
```

docs/source/articles/sequential-order-numbering.html.md
```diff
@@ -0,0 +1,77 @@
...
+end
+```
+
+However, Workarea's order IDs definitely don't look "numerical", and may be confusing to read for those used to relational systems. For this reason, you may want to customize how admins view orders on their site.
```
bencrouse, line 23:
```suggestion
However, Workarea's order IDs definitely don't look numerical, and may not be what people are used to. For this reason, merchants may insist on customizing how admins view orders on their site.
```

1) Let's not create the thought they might be confusing and 2) let's imply they are pushing back (which they should be). 😄

docs/source/articles/sequential-order-numbering.html.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Customizing Sequential Order Numbers
+
+Some clients may want this display ID to be sequentially numbered. This is possible to customize, but there's one caveat to keep in mind when doing so: MongoDB is a **distributed database**, so in production there are many instances of Mongo running to make sure that your app's data is not only within reach, but can be retrieved efficiently. This can lead to gaps in the order numbers, as well as order numbers appearing out of order in the admin (since Workarea only shows you orders which have been placed).
```
bencrouse, line 27: It seems like we should explain the caveats _after_ introducing the approach. They need to wrap their heads around it. Maybe before and after to try to drive the message home?

bencrouse (changes requested):

docs/source/articles/sequential-order-numbering.html.md
tubbo: It should actually be `assert(orders.second.id > orders.first.id)` :D

tubbo: this actually should be `assert(orders.second.id > orders.first.id)` :D

docs/source/articles/sequential-order-numbering.html.md
tubbo: I like the second part, but I'm not sure we're "creating" the thought here for merchants if they're asking for this in the first place. I will reword it, however, to drive home the point that a _merchant_ is the one who asks for this, not a developer.

docs/source/articles/sequential-order-numbering.html.md
tubbo: I'd like to do both before & after because I feel like the caveats are worth mentioning a few times.

docs/source/articles/sequential-order-numbering.html.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Sequential Order Numbering
+created_at: 2020/09/24
+excerpt: "In Workarea, orders (and other persisted data) are identified in a different manner from relational databases in order to avoid collisions. This guide explains how Workarea identifies orders, and how you might want to customize this for vanity purposes."
+---
+
+# Sequential Order Numbering
+
+In relational database systems, identifiers are typically numbers that are automatically incremented. Many ecommerce systems, both in the past and still running today, are built around relational databases in order to persist their data. As a result, order identifiers tend to have a "numerical" look to them, so much so that the colloqualism "order number" has been adopted throughout the industry. In Workarea, orders (and other persisted data) are identified in a different manner to avoid collisions when the backing database store is distributed amongst many machines. This guide explains how Workarea identifies orders, and how you might want to customize this for vanity purposes.
```
bencrouse, line 9:
```suggestion
In relational database systems, identifiers are typically numbers that are automatically incremented. Many ecommerce systems, both in the past and still running today, are built around relational databases. As a result, order identifiers tend be numerical, so much so that the colloqualism "order number" has been adopted throughout the industry. In Workarea, orders (and other persisted data) are identified in a different manner to avoid collisions when the backing database store is distributed amongst many machines. This guide explains how Workarea identifies orders, and how you might want to customize this for vanity purposes.
```

Unnecessary words, and "numerical" isn't a "look", it's either a number or it's not.

docs/source/articles/sequential-order-numbering.html.md
bencrouse: What I mean is they might not have been thinking about that and now by mentioning it they may go "hmm, _are_ these confusing?"

docs/source/articles/sequential-order-numbering.html.md
```diff
@@ -0,0 +1,77 @@
...
+
+## How Workarea Identifies Orders
+
+When an order is created or instantiated, it is assigned a randomized 5-digit uppercase hexadecimal value as its ID. Workarea does this so order numbers are legible to human eyes, and will play nicely on MongoDB. This ID is used throughout the system to link orders to their Payment and Shipping information, as well as in the admin for display purposes. Although this order ID is "randomized", `SecureRandom` generation uses the current system time to determine a seed value for its algorithm, so when compared, they are actually sequential already:
+
+```ruby
+def test_sequential_order_numbering
+  orders = [create_order, create_order]
+
+  assert(orders.second.id > orders.first) # => true
```
bencrouse, line 19: You probably meant: 
```suggestion
  assert(orders.second.id > orders.first.id) # => true
```

bencrouse (changes requested):
I think this could still be a little stronger on the empathy side... maybe something along the lines of this?

We understand many merchants and vendors in the ecommerce world are used to seeing numerical, incrementing order numbers. After using Workarea you may be wondering why we don't use these! They're everywhere!

Most of this legacy software is built on older, relational databases where in terms of performance, auto-incrementing numbers were the cheapest way to store IDs. This is fine and great, but this doesn't work out too well on modern distributed databases like MongoDB.

So you may get some pushback from merchants, looking for what they're used to. We've found explaining the rationale will usually help them get over this. But there are some merchants who are stuck on this point, and will insist on incrementing, numeric numbers for vanity purposes. This guide will show you the best way to achieve something we really don't encourage any project to do.

bencrouse (approved):

PR #521, opened 2020-09-23
Notate Rack::Cache Ignore Params In Documentation

tubbo:
Add documentation explaining why one might want to ignore query string params in `Rack::Cache`, how to do it in a Workarea environment, and which plugins will do it for you.

### DIFF
--- a/docs/source/articles/http-caching.html.md
+++ b/docs/source/articles/http-caching.html.md
@@ -31,7 +31,31 @@ This creates a major complication for any data on the page with user or session 
 
 **Note:** [Rack::Cache](http://rtomayko.github.io/rack-cache/) is a simple way to achieve the caching advantage explained above without the configuration and hosting burden of a more robust HTTP cache like [Varnish](https://www.varnish-cache.org).
 
-## Customizing the Rack::Cache Key
+## Using Rack::Cache
+
+Many Workarea applications use `Rack::Cache` for storing their HTTP cache. In fact, this is the caching solution that is used by [Workarea Commerce Cloud](https://www.workarea.com/pages/commerce-cloud), so if you're on Commerce Cloud, congratulations! You're already using `Rack::Cache`! 🎉
+
+Now that we're done celebrating, we can focus on a few key elements of `Rack::Cache` that are configurable and notable for Workarea applications.
+
+### Ignore User-Specific Tracking Parameters in the Query String
+
+Many stores on the Workarea platform use marketing tools suck as [Listrak](https://github.com/workarea-commerce/workarea-listrak) or [Emarsys](https://github.com/workarea-commerce/workarea-emarsys) to handle abandoned cart emails, marketing campaigns, and so on. These services tend to use query parameters within the URL to identify each individual user, and this can cause an issue with caching. `Rack::Cache` will, by default, generate cache keys based on the given URL. So, if you have multiple requests to your **/categories/mens** page, like this:
+
+```
+GET /categories/mens?tracking_id=96a30500-6add-47e1-9ee8-cf3b5052ecf3
+GET /categories/mens?tracking_id=9d9948e3-df0b-4b17-9396-4bbcbac7f4c9
+GET /categories/mens?tracking_id=0ffc0e09-4041-4876-8f8e-05ae87bc3bf3
+```
+
+That will result in multiple entries written to the cache for the same exact request. This is inefficient at best, and dangerous to your performance at worst. To make sure you won't run into this situation, `Rack::Cache` [allows you to configure ignored query parameters](https://github.com/rtomayko/rack-cache#ignoring-tracking-parameters-in-cache-keys), and since Workarea uses its own subclass of `Rack::Cache::Key`, you'll need to customize the tracking params on that class like so:
+
+```ruby
+Workarea::Cache::RackCacheKey.query_string_ignore = proc { |key, value| key == 'tracking_id' }
+```
+
+If you're using the `Workarea::Listrak` plugin, you're in luck! 🍀 Recent versions of the plugin [include configuration for the cache key](https://github.com/workarea-commerce/workarea-listrak/blob/master/config/initializers/rack_cache.rb) out of the box, so you won't need to do anything to get efficient caches in production.
+
+### Customizing the Cache Key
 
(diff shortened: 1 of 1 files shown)


docs/source/articles/http-caching.html.md
```diff
@@ -31,7 +31,31 @@ This creates a major complication for any data on the page with user or session
 
 **Note:** [Rack::Cache](http://rtomayko.github.io/rack-cache/) is a simple way to achieve the caching advantage explained above without the configuration and hosting burden of a more robust HTTP cache like [Varnish](https://www.varnish-cache.org).
 
-## Customizing the Rack::Cache Key
+## Using Rack::Cache
+
+Many Workarea applications use `Rack::Cache` for storing their HTTP cache. In fact, this is the caching solution that is used by [Workarea Commerce Cloud](https://www.workarea.com/pages/commerce-cloud), so if you're on Commerce Cloud, congratulations! You're already using `Rack::Cache`! 🎉
```
bencrouse, line 36: This section doesn't really make sense to me - all Workarea applications use `Rack::Cache`. You would have to go out of your way to decorate it out.

docs/source/articles/http-caching.html.md
```diff
@@ -31,7 +31,31 @@ This creates a major complication for any data on the page with user or session
...
+
+Now that we're done celebrating, we can focus on a few key elements of `Rack::Cache` that are configurable and notable for Workarea applications.
+
+### Ignore User-Specific Tracking Parameters in the Query String
+
+Many stores on the Workarea platform use marketing tools suck as [Listrak](https://github.com/workarea-commerce/workarea-listrak) or [Emarsys](https://github.com/workarea-commerce/workarea-emarsys) to handle abandoned cart emails, marketing campaigns, and so on. These services tend to use query parameters within the URL to identify each individual user, and this can cause an issue with caching. `Rack::Cache` will, by default, generate cache keys based on the given URL. So, if you have multiple requests to your **/categories/mens** page, like this:
```
bencrouse, line 42:
```suggestion
Many stores on the Workarea platform use marketing tools suck as [Listrak](https://github.com/workarea-commerce/workarea-listrak) or [Emarsys](https://github.com/workarea-commerce/workarea-emarsys) to handle abandoned cart emails, marketing campaigns, and so on. These services sometimes use query parameters within the URL to identify each individual user, and this can cause an issue with caching. `Rack::Cache` will, by default, generate cache keys based on the given URL. So, if you have multiple requests to your **/categories/shirts** page, like this:
```

docs/source/articles/http-caching.html.md
```diff
@@ -31,7 +31,31 @@ This creates a major complication for any data on the page with user or session
...
+
+```
+GET /categories/mens?tracking_id=96a30500-6add-47e1-9ee8-cf3b5052ecf3
```
bencrouse, line 45: Let's try to be sensitive since there's no need to use gender-specific URLs.

docs/source/articles/http-caching.html.md
```diff
@@ -31,7 +31,31 @@ This creates a major complication for any data on the page with user or session
...
+GET /categories/mens?tracking_id=9d9948e3-df0b-4b17-9396-4bbcbac7f4c9
+GET /categories/mens?tracking_id=0ffc0e09-4041-4876-8f8e-05ae87bc3bf3
+```
+
+That will result in multiple entries written to the cache for the same exact request. This is inefficient at best, and dangerous to your performance at worst. To make sure you won't run into this situation, `Rack::Cache` [allows you to configure ignored query parameters](https://github.com/rtomayko/rack-cache#ignoring-tracking-parameters-in-cache-keys), and since Workarea uses its own subclass of `Rack::Cache::Key`, you'll need to customize the tracking params on that class like so:
+
+```ruby
+Workarea::Cache::RackCacheKey.query_string_ignore = proc { |key, value| key == 'tracking_id' }
```
bencrouse, line 53: Is this really true? Since it subclasses `Rack::Cache::Key`, I'd expect it to work with the out-of-the-box class.

bencrouse (changes requested):

docs/source/articles/http-caching.html.md
tubbo: That's true. This doc sorta reads like you have a choice of using `Rack::Cache` vs. Varnish, and I would definitely like to rewrite some parts to emphasize that you get rack-cache configured in prod by default.

docs/source/articles/http-caching.html.md
bencrouse: Update - didn't realize we were using this elsewhere already

docs/source/articles/http-caching.html.md
tubbo: Yes.

```ruby
require 'rack/cache/key'

class Key < Rack::Cache::Key; end

Rack::Cache::Key.query_string_ignore = 'foo'
Key.query_string_ignore # => nil
```

docs/source/articles/http-caching.html.md
```diff
@@ -29,9 +28,29 @@ Controller/actions which enable this include:
 
 This creates a major complication for any data on the page with user or session specific. Examples include login status, cart count, and personalized product recommendations. Workarea solves this problem by loading this content asynchronously via Javascript. Make sure you review any customized functionality to ensure you won't be caching pages with session-specific data.
 
-**Note:** [Rack::Cache](http://rtomayko.github.io/rack-cache/) is a simple way to achieve the caching advantage explained above without the configuration and hosting burden of a more robust HTTP cache like [Varnish](https://www.varnish-cache.org).
+## Using Rack::Cache
+
+Workarea applications use [Rack::Cache](https://github.com/rtomayko/rack-cache) to store their HTTP cache. `Rack::Cache` is an HTTP caching Rack middleware that is configured in qa, staging, and production environments. It's backed by [Redis](https://github.com/redis-store/redis-rack-cache), in which Rack::Cache stores all of its metadata and actual cached content.
```
bencrouse, line 33:
```suggestion
Workarea applications use [Rack::Cache](https://github.com/rtomayko/rack-cache) to store their HTTP cache. `Rack::Cache` is a HTTP caching Rack middleware. It's backed by [Redis](https://github.com/redis-store/redis-rack-cache), in which `Rack::Cache` stores all of its metadata and actual cached content.

Out of the box, the effects of `Rack::Cache` are disabled in `test` and `development` environments.
```

bencrouse (approved):

PR #524, opened 2020-09-24
Improve UX of default search filter sizes

bencrouse:


tubbo (approved):

core/config/initializers/00_configuration.rb
```diff
@@ -127,11 +127,6 @@
   end
 
   fieldset 'Search', namespaced: false do
-    field 'Default Search Facet Result Sizes',
-      type: :integer,
-      default: 10,
-      description: 'The number of filter results returned for each filter type.'
-
     field 'Search Facet Result Sizes',
```
mttdffy, line 130: the description of this now doesn't make sense since it references the "default config above"

mttdffy (changes requested):

core/config/initializers/00_configuration.rb
bencrouse: In the code "above" refers to the specific filter settings, and it renders that way in the UI too
Screen Shot 2020-09-28 at 8 19 05 AM

mttdffy: Im talking about "search facet result sizes" description, says "the default config above", which is now below it.

bencrouse: Oh derp, I've updated.

PR #525, opened 2020-09-24
Remove unnecessary Capybara blocking when testing content is not present

jonmast:
Capybara's `page.has_content?` and similar methods block until a timeout
is reached if they can't find the content. This is not what we want if
we're checking that the content does *not* exist, switch to using
`refute_text` in these scenarios.

The timeout is equal to the number of installed plugins and we have
client apps with 30+, which means that the 38 instances removed in this
commit could represent twenty minutes of unnecessary blocking in some
scenarios.

Thanks to James Anaipakos in
https://discourse.workarea.com/t/capybara-refute-assertions-waiting-full-default-wait-time/1610
for alerting me to the issue.

Note: I've selected `refute_text` and `refute_selector` somewhat arbitrarily
here, they have several aliases. I'm specifically avoiding the more closely
analogous `page.has_no_content?` because the resulting error messages
aren't as nice, but I'm happy to update to whatever you all prefer.

### DIFF
--- a/admin/test/system/workarea/admin/assets_system_test.rb
+++ b/admin/test/system/workarea/admin/assets_system_test.rb
@@ -27,7 +27,7 @@ module Workarea
 
         assert_current_path(admin.content_assets_path)
         assert(page.has_content?('Success'))
-        refute(page.has_content?('Edited Asset'))
+        refute_text('Edited Asset')
       end
 
       def test_insertion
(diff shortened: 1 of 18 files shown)


bencrouse:
Nice! Thanks for opening @jonmast . 

I'd like to see a passing build on this before merging. One reason it might not have run is we have secrets set up, so allowing forks to run builds would allow listing secrets out. So, I'm going to cherry pick this into a branch on our repo. And we can move the discussion/review there.

Cheers!

bencrouse:
Merged in #526

PR #528, opened 2020-09-25
Fix safe navigation method calls

bencrouse:
This will raise if the menu content is nil.


bencrouse:
Discovered during site builder testing.

tubbo (approved):

PR #532, opened 2020-09-30
Fix skip services

bencrouse:
This was broken due to the admin-based configuration looking for a Mongo
connection.

### DIFF
--- a/core/lib/workarea.rb
+++ b/core/lib/workarea.rb
@@ -182,6 +182,16 @@ module Workarea
   def self.deprecation
     @deprecation ||= ActiveSupport::Deprecation.new('3.6', 'Workarea')
   end
+
+  # Whether the app should skip connecting to external services on boot,
+  # such as Mongo, Elasticsearch, or Redis. Note that this will break
+  # functionality relying on these services.
+  #
+  # @return [Boolean]
+  #
+  def self.skip_services?
+    !!(ENV['WORKAREA_SKIP_SERVICES'] =~ /true/)
+  end
 end
 
 require 'workarea/core'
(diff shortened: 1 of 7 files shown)


core/lib/workarea.rb
```diff
@@ -182,6 +182,16 @@ def self.define_content_block_types(&block)
...
   end
+
+  # Whether the app should skip connecting to external services on boot,
+  # such as Mongo, Elasticsearch, or Redis. Note that this will break
+  # functionality relying on these services.
+  #
+  # @return [Boolean]
+  #
+  def self.skip_services?
+    ENV['WORKAREA_SKIP_SERVICES'] =~ /true/
```
mttdffy, line 193: probably not relevant here, but I've recently discovered that while the `to_s` is not necessary for this type of condition when you are concerned about `nil`, it does result in a `true` boolean not returning `true`

```
true =~ /true/i #=> nil
```

mttdffy (approved):

core/lib/workarea.rb
tubbo: i really don't think we should be so strict in terms of the value for this env var. not having the var set should result in connecting to services. having the var set to anything should result in not connecting to services. this is how `$RAILS_LOG_TO_STDOUT` and `$RAILS_SERVE_STATIC_FILES` works, afaik. you can set them to any value, and as long as they're set, their config takes effect.

tubbo: i really don't think we should be so strict in terms of the value for this env var. not having the var set should result in connecting to services. having the var set to anything should result in not connecting to services. this is how $RAILS_LOG_TO_STDOUT and $RAILS_SERVE_STATIC_FILES works, afaik. you can set them to any value, and as long as they're set, their config takes effect.

tubbo (changes requested):

PR #533, opened 2020-10-06
Patch RefererParser for Android URLs

### LINKED ISSUE #531: RefererParser::InvalidUriError (issue for builds from 3.4.0-beta1 to 3.5.0-beta1)
Edit: We probably want to label this not as a `bug` but more as `enhancement` (my bad)

**Describe the bug**
Sentry is reporting an error for one of our build:

image

**Investigation**

- RefererParser is a dependency in workarea-core: `s.add_dependency 'referer-parser', '~> 0.3.0' `(https://rubygems.org/gems/referer-parser/versions/0.3.0) and seems to have been introduced in workarea `3.4.0-beta1`

- Some users of this gem are experiencing similar problems: https://github.com/snowplow-referer-parser/referer-parser/issues/131. The gist being that `RefererParser::Parser.new.parse()` is not able to parse referrer "urls" starting with `android-app://` (ie. The referrer being an android app such as slack, facebook, twitter, ...).

-  *Scott Bullard* (can't link his github profile) provided a fix for this bug: https://github.com/snowplow-referer-parser/referer-parser/pull/145/commits/737a892f628347a610e5633d62153c8d6518f4e5, however it has never been officially released (merged on `develop` but not on `master`)

- Looking at the different versions of Workarea, it seems that 3.5.0 mitigates the bug by rescue-ing the parse result: https://github.com/workarea-commerce/workarea/blob/v3.5.0/core/lib/workarea/visit.rb#L69. However that's not the case for older versions starting from `3.4.0-beta1` to `3.5.0-beta1`: https://github.com/workarea-commerce/workarea/blob/v3.4.12/core/app/controllers/workarea/current_referrer.rb#L10 (I couldn't fetch the source code from 3.4.0-beta1 `¯\_(ツ)_/¯ `)

- It's worth noting that recent builds (>= v3.5.0) will basically not track the android-app referrers since the gem `RefererParser` will raise and therefore will be rescued to an empty hash. It's probably worth fixing since it's a good metric to have (that's what triggered this issue / discussion).

**Bonus**

If anyone is running into the same issue and are not able to upgrade at the moment, here's my fix (v3.4.32):

```ruby
# lib/workarea/ext/freedom_patches/referer_parser.rb
module RefererParser
  class Parser
    #
    # Patched to take in account the new scheme `android-app`.
    #
    # See: https://github.com/snowplow-referer-parser/referer-parser/pull/145/commits/737a892f628347a610e5633d62153c8d6518f4e5
    #
    # The PR above has never been officialy released, therefore the patch.
    #
    def parse(obj)
      url = obj.is_a?(URI) ? obj : URI.parse(obj.to_s)
      
      # This section changed
      if !['android-app', 'http', 'https'].include?(url.scheme)
        raise InvalidUriError.new("Only Android-App, HTTP and HTTPS schemes are supported -- #{url.scheme}")
      end

      data = { :known => false, :uri => url.to_s }

      domain, name_key = domain_and_name_key_for(url)
      if domain and name_key
        referer_data = @name_hash[name_key]
        data[:known] = true
        data[:source] = referer_data[:source]
        data[:medium] = referer_data[:medium]
        data[:domain] = domain

        # Parse parameters if the referer uses them
        if url.query and referer_data[:parameters]
          query_params = CGI.parse(url.query)
          referer_data[:parameters].each do |param|
            # If there is a matching parameter, get the first non-blank value
            if !(values = query_params[param]).empty?
              data[:term] = values.select { |v| v.strip != "" }.first
              break if data[:term]
            end
          end
        end
      end

      data
    rescue URI::InvalidURIError
      raise InvalidUriError.new("Unable to parse URI, not a URI? -- #{obj.inspect}", $!)
    end
  end
end
```

```ruby
# test/lib/workarea/ext/freedom_patches/referer_parser_test.rb
# A silly test to sleep well at night
require 'test_helper'

module RefererParser
  class ParserTest < Workarea::TestCase
    def test_parse_android_app_referrers
      referrers = [
        'android-app://com.linkedin.android',
        'android-app://org.telegram.plus',
        'android-app://com.twitter.android'
      ]

      assert_nothing_raised do
        referrers.each do |referrer|
          Parser.new.parse(referrer)
        end
      end
    end
  end
end
```

```ruby
# You probably want to override Workarea::CurrentReferrer as well to backport the rescue
# app/controllers/workarea/current_referrer.rb
module Workarea
  module CurrentReferrer
    #
    # Base implementation (workarea-core v3.4.32) is prone
    # to raise an error for `RefererParser::Parser.new.parse(referrer)`
    # which will block the user to be able to checkout properly.
    #
    # @source https://github.com/workarea-commerce/workarea/blob/v3.4.32/core/app/controllers/workarea/current_referrer.rb
    #
    # The latest implementation (workarea-core 3.5.18) mitigates the problem
    # by rescue-ing the attributes.
    #
    # @source https://github.com/workarea-commerce/workarea/blob/v3.5.18/core/lib/workarea/visit.rb#L72
    #
    # I took inspiration from the latest implementation and applied the changes here.
    #
    def current_referrer
      return @current_referrer if defined?(@current_referrer)

      referrer = cookies['workarea_referrer']
      return unless referrer.present?
     
      # This rescue is gold, you don't want the user to reach a 404 during checkout ===> -$$$$
      attributes = RefererParser::Parser.new.parse(referrer) rescue {}

      @current_referrer ||= TrafficReferrer.new(
        attributes.slice(:source, :medium, :uri)
      )
    end
  end
end
```

tubbo:
Android App URLs have a special android-app:// scheme that is rejected by the currently released version of the referer-parser gem. The code in this patch already exists in the master branch of the gem, but this has not yet been released, and if Android users browse the storefront it can generate an error when collecting referer information.

Fixes #531

### DIFF
--- a/core/app/controllers/workarea/current_referrer.rb
+++ b/core/app/controllers/workarea/current_referrer.rb
@@ -9,6 +9,7 @@ module Workarea
       @current_referrer ||= TrafficReferrer.new(
         RefererParser::Parser.new.parse(referrer).slice(:source, :medium, :uri)
       )
+    rescue RefererParser::InvalidUriError
     end
   end
 end
--- a/core/lib/workarea/core.rb
+++ b/core/lib/workarea/core.rb
@@ -140,6 +140,7 @@ require 'workarea/ext/mongoid/find_ordered'
 require 'workarea/ext/sprockets/ruby_processor'
 require 'workarea/ext/jbuilder/jbuilder_append_partials'
 require 'workarea/ext/jbuilder/jbuilder_cache'
+require 'workarea/ext/referer_parser/parser.decorator'
 
 if Rails.env.development?
   require 'workarea/ext/freedom_patches/routes_reloader'
--- a/core/lib/workarea/ext/referer_parser/parser.decorator
+++ b/core/lib/workarea/ext/referer_parser/parser.decorator
@@ -1,0 +1,43 @@
+module RefererParser
+  # This code is actually in `master` branch of
+  # https://github.com/snowplow-referer-parser/ruby-referer-parser but
+  # has not yet been released. It's used to allow through android-app://
+  # URLs.
+  decorate Parser do
+    # Given a string or URI, return a hash of data
+    def parse(obj)
+      url = obj.is_a?(URI) ? obj : URI.parse(obj.to_s)
+
+      unless ['android-app', 'http', 'https'].include?(url.scheme)
+        raise InvalidUriError, "Only Android-App, HTTP, and HTTPS schemes are supported -- #{url.scheme}"
+      end
+
+      data = { known: false, uri: url.to_s }
+
+      domain, name_key = domain_and_name_key_for(url)
+      if domain && name_key
+        referer_data = @name_hash[name_key]
+        data[:known] = true
+        data[:source] = referer_data[:source]
+        data[:medium] = referer_data[:medium]
+        data[:domain] = domain
+
+        # Parse parameters if the referer uses them
+        if url.query && referer_data[:parameters]
+          query_params = CGI.parse(url.query)
+          referer_data[:parameters].each do |param|
+            # If there is a matching parameter, get the first non-blank value
+            unless (values = query_params[param]).empty?
+              data[:term] = values.reject { |v| v.strip == '' }.first
+              break if data[:term]
+            end
+          end
+        end
+      end
+
+      data
+    rescue URI::InvalidURIError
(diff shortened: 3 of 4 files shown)


tubbo:
I filed an issue at snowplow-referer-parser/ruby-referer-parser#10 to get a new gem release going.

bencrouse (changes requested):
Any reason you didn't go with the rescue that @GesJeremie suggested? Seems more bulletproof... we would never want anyone's checkout flow to be interrupted due to failed referrer parsing.

bencrouse (approved):

PR #536, opened 2020-10-07
Fix Test That Will Never Fail

tubbo:
This test for the `StatusReporter` worker asserted `2`, which will never fail because `2` will never be falsy. Updated the assertion to use the intended `assert_equals`


bencrouse (approved):

bencrouse:
Nice

PR #537, opened 2020-10-08
Refactor product entries to allow accessing logic per-product

bencrouse:
This allows easier reuse of this logic, specifically for the site
builder plugin we're working on.


mttdffy (approved):
🤮  When did this class get so messy. I regret not making this class only accept a single product, with maybe a method to loop through products and create an instance for each product. Feel like that would alleviate some of this mess but probably not doable for a minor. SAD.

tubbo (approved):

PR #539, opened 2020-10-12
Prevent Clearing Out Navigable When Saving Taxons

### LINKED ISSUE #534: Saving taxon clears navigable_id field
Editing a taxon causes the navigable_id param field to pass an empty value, losing the data on save. 

**To Reproduce**
1. Go to /admin/navigation_taxons
2. Select any taxon to edit (I've tested this with category taxons)
3. Save the taxon.
4. Navigated back to the taxon you just edited. 
5. Notice the "name" in the links to taxon is blank. 

**Expected behavior**
A taxon should retain its selected values. 

 - Workarea Version: workarea-3.5.20
 - Verified this on Workarea Demo setup from instructions here: https://github.com/workarea-commerce/workarea/blob/master/demo/README.md

tubbo:
The `WORKAREA.newNavigationTaxons` module was looking in the wrong place for the selected navigable item, therefore the `selected` var would always return `undefined`, causing the `navigable_id` param to be blank every time. Fix this by querying for the correct DOM node (the `[data-new-navigation-taxon]` element) and pulling the selected ID from its data.

Fixes #534

### DIFF
--- a/admin/test/javascripts/fixtures/existing_navigation_taxon.html.haml
+++ b/admin/test/javascripts/fixtures/existing_navigation_taxon.html.haml
@@ -1,0 +1,10 @@
+.new-navigation-link__section{ data: { new_navigation_taxon: 'bar' } }
+  %fieldset
+    %legend
+      %span.heading.heading--2 Existing
+    .property
+      = label_tag 'navigable_type', 'Type', class: 'property__name'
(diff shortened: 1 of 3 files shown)


admin/test/javascripts/fixtures/existing_navigation_taxon.html.haml
```diff
@@ -0,0 +1,10 @@
+.new-navigation-link__section{ data: { new_navigation_taxon: 'bar' } }
```
tubbo, line 1: this is a copy of `new_navigation_taxon.html.haml`, except it mocks out an edit page where the `navigable_id` has already been selected.

bencrouse (approved):

PR #540, opened 2020-10-12
Add Multiplier for Bulk Editing Product Pricing

tubbo:
When bulk editing products, Workarea previously allowed the admin to set pricing on all SKUs in bulk. This was always a flat amount, and all prices were updated to match that amount. To build upon this feature, bulk editing products will now enable admins to set a percentage or amount they wish to increase/decrease the prices by. This allows, for example, a merchandiser to mark up all products in their bulk edit query 20%, or mark down all products by $15. The existing functionality for setting these prices has also been preserved, and extended to allow for setting the price to a percentage of the existing price, such as setting the price of a $20 to 20%, which would cause the product's price to be $4.

**Edit:**
Screen Shot 2020-10-15 at 12 11 35 PM

**Review:**
Screen Shot 2020-10-15 at 12 10 20 PM

### DIFF
--- a/admin/app/view_models/workarea/admin/bulk_action_product_edit_view_model.rb
+++ b/admin/app/view_models/workarea/admin/bulk_action_product_edit_view_model.rb
@@ -18,12 +18,42 @@ module Workarea
       end
 
       def pricing_prices
-        pricing.fetch('prices', []).first || {}
+        pricing['prices'] || {}
       end
 
       def segments
         return [] if settings['active_segment_ids'].blank?
         @segments ||= Segment.in(id: settings['active_segment_ids']).to_a
+      end
+
+      def pricing_actions
+        [
+          [
+            t('workarea.admin.bulk_action_product_edits.pricing.set'),
+            'set'
+          ],
+          [
+            t('workarea.admin.bulk_action_product_edits.pricing.increase'),
+            'increase'
+          ],
+          [
+            t('workarea.admin.bulk_action_product_edits.pricing.decrease'),
+            'decrease'
+          ]
+        ]
+      end
+
+      def pricing_types
+        [
+          [
+            Money.default_currency.symbol,
+            'flat'
+          ],
+          [
+            t('workarea.admin.bulk_action_product_edits.pricing.percentage'),
+            'percentage'
+          ]
+        ]
       end
     end
   end
--- a/admin/app/views/workarea/admin/bulk_action_product_edits/review.html.haml
+++ b/admin/app/views/workarea/admin/bulk_action_product_edits/review.html.haml
@@ -77,8 +77,20 @@
                   - @bulk_action.pricing.except('prices').each do |name, value|
                     %p <strong>#{t("workarea.admin.fields.#{name}", default: name.titleize)}</strong>: #{bulk_actions_display_value_for(value)}
                   - if @bulk_action.pricing['prices'].present?
-                    - @bulk_action.pricing['prices'].first.each do |name, value|
-                      %p <strong>#{t("workarea.admin.fields.#{name}", default: name.titleize)}</strong>: #{value}
+                    - @bulk_action.pricing['prices'].each do |name, value|
+                      - action = value['action']
+                      - type = value['type']
+                      - amount = value['amount']
+                      - kind = t("workarea.admin.bulk_action_product_edits.review.pricing.#{name}")
+
+                      - if amount.present?
+                        %p
+                          = t("workarea.admin.bulk_action_product_edits.review.pricing.#{action}_html", kind: kind)
+                          %strong
+                            - if type == 'flat'
+                              = number_to_currency(amount)
+                            - else
+                              = number_to_percentage(amount, strip_insignificant_zeros: true)
                 - else
                   %p= t('workarea.admin.bulk_action_product_edits.review.no_change')
 
--- a/admin/config/locales/en.yml
+++ b/admin/config/locales/en.yml
@@ -200,7 +200,15 @@ en:
           settings: SKU Pricing Settings
           title: Pricing
           info: Prices Info
-          info_message: For products with existing prices, regular and sale prices will be applied to all prices with a minimum quantity of 1.
+          info_message: >
+            For products with existing prices, regular and sale prices
+            will be applied to all prices with a minimum quantity of 1.
+            Setting all prices to a flat amount does not apply to tiered
+            pricing.
+          increase: 'Increase all prices by'
+          decrease: 'Decrease all prices by'
+          set: 'Set prices with a minimum quantity of 1 to'
+          percentage: '%'
         review:
           no_change: No change
           save_and_finish: Save and Finish
@@ -218,6 +226,12 @@ en:
           with_a_new_release: With a new release
           with_release: With %{release_name}
           which_release_placeholder: Mother's Day %{year}
+          pricing:
+            increase_html: 'Increase <strong>all %{kind} prices</strong> by'
+            decrease_html: 'Decrease <strong>all %{kind} prices</strong> by'
+            set_html: 'Set <strong>%{kind} prices</strong> with a minimum quantity of 1 to'
+            regular: regular
+            sale: sale
         settings:
           title: Settings
         tags:
--- a/admin/test/system/workarea/admin/bulk_actions_system_test.rb
+++ b/admin/test/system/workarea/admin/bulk_actions_system_test.rb
@@ -4,6 +4,7 @@ module Workarea
   module Admin
     class BulkActionsSystemTest < SystemTest
       include Admin::IntegrationTest
+      include ActionView::Helpers::SanitizeHelper
 
       setup :create_products
       setup :set_config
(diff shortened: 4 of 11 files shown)


core/test/models/workarea/bulk_action/product_edit_test.rb
```diff
@@ -52,7 +52,16 @@ def test_updating_pricing_skus
         product_2 = create_product(variants: [{ sku: 'SKU3', regular: 15.00 }])
 
         edit = ProductEdit.new(
-          pricing: { msrp: '19.99', prices: [{ regular: '14.99' }] }
```
tubbo, line 55: One big difference here is that I've changed the `prices:` param to be a Hash rather than an Array. I took a look back into the history of this code, and it seems like the Array of prices was added in the first WIP implementation. We started off just trying to pull off an `accepts_nested_attributes_for :prices` thing, but that was not being set correctly in releases, so Duffy made a change that explicitly ran `price.update!` on the given params for each price in each SKU. By not changing the type, this meant that he didn't need to make any breaking changes to the form, which was definitely the right choice at the time. Now that we need to change the form anyway, I felt like there was no need keeping around this old API. It's difficult to work with, and makes sense in the modern bulk edit workflow. Changing the type to a Hash makes the resulting `apply_pricing!` code a lot cleaner and easier to read.

tubbo:
We changed some of the UI after the UX meeting...

## Old

**Edit:**
Screen Shot 2020-10-12 at 5 46 44 PM

**Review:**
Screen Shot 2020-10-12 at 5 47 01 PM

## New

**Edit:**
Screen Shot 2020-10-15 at 12 11 35 PM

**Review:**
Screen Shot 2020-10-15 at 12 10 20 PM

admin/app/view_models/workarea/admin/bulk_action_product_edit_view_model.rb
```diff
@@ -18,13 +18,43 @@ def selected_false?(hash, field)
...
+            'increase'
+          ],
+          [
+            t('workarea.admin.bulk_action_product_edits.pricing.decrease'),
+            'decrease'
+          ]
+        ]
+      end
+
+      def pricing_types(currency)
```
bencrouse, line 46: Why pass in currency here, isn't it always going to be `Money.default_currency`?

admin/app/views/workarea/admin/bulk_action_product_edits/review.html.haml
```diff
@@ -77,8 +77,20 @@
...
+                      - action = value['action']
+                      - type = value['type']
+                      - amount = value['amount']
+                      - kind = t("workarea.admin.bulk_action_product_edits.review.pricing.#{name}")
+
+                      - if amount.present?
+                        %p
+                          = t("workarea.admin.bulk_action_product_edits.review.pricing.#{action}_html", kind: kind)
+                          %strong
+                            - if value['type'] == 'flat'
```
bencrouse, line 90: This is in a variable above, might as well use that.

admin/app/views/workarea/admin/bulk_action_product_edits/review.html.haml
```diff
@@ -77,8 +77,20 @@
                   - @bulk_action.pricing.except('prices').each do |name, value|
                     %p <strong>#{t("workarea.admin.fields.#{name}", default: name.titleize)}</strong>: #{bulk_actions_display_value_for(value)}
                   - if @bulk_action.pricing['prices'].present?
-                    - @bulk_action.pricing['prices'].first.each do |name, value|
-                      %p <strong>#{t("workarea.admin.fields.#{name}", default: name.titleize)}</strong>: #{value}
+                    - @bulk_action.pricing['prices'].each do |name, value|
+                      - action = value['action']
```
bencrouse, line 81: Do these make sense as a variable if it's used in one spot?

admin/config/locales/en.yml
```diff
@@ -200,7 +200,15 @@ en:
           settings: SKU Pricing Settings
           title: Pricing
           info: Prices Info
-          info_message: For products with existing prices, regular and sale prices will be applied to all prices with a minimum quantity of 1.
+          info_message: >
+            For products with existing prices, regular and sale prices
+            will be applied to all prices with a minimum quantity of 1.
+            Sites will be
```
bencrouse, line 206: "Sites will be"?

admin/config/locales/en.yml
```diff
@@ -200,7 +200,15 @@ en:
...
+          increase: 'Increase all prices by'
+          decrease: 'Decrease all prices by'
+          set: 'Set prices with a minimum quantity of 1 to'
+          flat: '%{currency}'
```
bencrouse, line 210: This seems a little odd, why not output the currency?

(edit) maybe for dynamic reasons?

admin/test/system/workarea/admin/bulk_actions_system_test.rb
```diff
@@ -194,7 +202,15 @@ def test_bulk_editing
...
-        assert(page.has_content?("#{t('workarea.admin.fields.regular')}: 20"))
+        assert_text(
+          strip_tags(
+            t(
+              'workarea.admin.bulk_action_product_edits.review.pricing.set_html',
+              kind: t('workarea.admin.bulk_action_product_edits.review.pricing.regular')
+            )
+          )
+        )
+        assert_text('$20.00')
```
bencrouse, line 213: This will break if the default currency is changed.

core/app/models/workarea/bulk_action/product_edit.rb
```diff
@@ -52,13 +58,27 @@ def apply_pricing!(product)
...
+                next unless value['amount'].present?
+
+                if price.generic? || change_non_generic_price?(value)
+                  params[type] = amount_from(value, price[type])
+                end
+              end
+
+              price.update_attributes!(updates)
             end
+          elsif prices.present?
```
bencrouse, line 76: I'm not sure this makes sense, if you're adjusting by a percentage I think you'd be pretty surprised a new price gets created. Let's remove this.

core/app/models/workarea/bulk_action/product_edit.rb
```diff
@@ -52,13 +58,27 @@ def apply_pricing!(product)
...
-          prices = sku_pricing.delete('prices') unless sku.prices.blank?
+          prices = sku_pricing.delete('prices')
           sku.update_attributes!(sku_pricing)
 
-          if prices.present?
-            sku.prices.select(&:generic?).each do |price|
-              price.update_attributes!(prices.first)
+          if prices.present? && sku.prices.present?
+            sku.prices.each do |price|
+              updates = prices.each_with_object({}) do |(type, value), params|
```
bencrouse, line 66: This code feels like there's a missing class/object due to all the methods written into this class to support it. Try to get it to something like this maybe:
```
sku.prices.each do |price|
  update = PriceUpdate.new(price, prices)
  price.update!(update.attributes) if update.apply?
end
```

bencrouse (changes requested):

core/app/models/workarea/bulk_action/product_edit.rb
tubbo: yeah I was thinking about adding a new class to encapsulate all this logic.

admin/config/locales/en.yml
tubbo: yeah i was thinking if someone wanted to output something like "$ (USD)" vs "$ (AUD)" rather than "$", on different locales.

admin/app/views/workarea/admin/bulk_action_product_edits/review.html.haml
tubbo: It doesn't matter much to me, I probably did it to keep the text width down and make the code have less brackets.

core/app/models/workarea/bulk_action/product_edit/price_change/amount.rb
```diff
@@ -0,0 +1,80 @@
...
+            @amount = amount
+          end
+
+          # Test whether this amount is not setting a flat amount. Used
+          # in the `PriceChange` class when determining whether to
+          # perform an update. Non-generic prices are not updated when
+          # the admin chooses to set a flat amount for all prices in the
+          # product's SKUs.
+          #
+          # @return [Boolean] Whether the admin wants to set a flat amount.
```
bencrouse, line 33: This comment is incorrect, right?

core/app/models/workarea/bulk_action/product_edit/price_change/amount.rb
```diff
@@ -0,0 +1,80 @@
...
+          def multiplier?
+            return true if action != 'set'
+
+            type != 'flat'
```
bencrouse, line 37: These 2 `!=` are kind of confusing, aren't there specific values to compare against to make it a little clearer?

core/app/models/workarea/bulk_action/product_edit/price_change/amount.rb
```diff
@@ -0,0 +1,80 @@
...
+            end
+          end
+
+          private
+
+          # Calculate the percentage of a given price.
+          #
+          # @private
+          # @return [Float] Percentage of the current price.
+          def percent
```
bencrouse, line 73:
```suggestion
          def applied_percent
```

bencrouse (changes requested):
Looking much better!

core/app/models/workarea/bulk_action/product_edit/price_change/amount.rb
tubbo: Not that I can tell. A price change is a "multiplier" if you're not trying to set a flat amount.

bencrouse (changes requested):
Some test failures in the build...

bencrouse (approved):

PR #542, opened 2020-10-14
Merge metrics when a user's email is updated

bencrouse:
This ensures the old metrics info stays around after the email change.

WORKAREA-294


core/app/models/workarea/metrics/user.rb
```diff
@@ -123,6 +123,30 @@ def average_order_value
         return nil if orders.zero?
         revenue / orders.to_f
       end
+
+      def merge!(other)
```
bencrouse, line 127: This is ported from v3.5.

core/app/workers/workarea/update_email.rb
```diff
@@ -0,0 +1,33 @@
+module Workarea
+  class UpdateEmail
```
bencrouse, line 2: I checked, this wasn't being decorated anywhere I could find and I wanted to keep the number of Sidekiq jobs at a minimum.

tubbo (approved):

PR #544, opened 2020-10-16
Add Note To Category Default Sort Edit

tubbo:
The selected `default_sort` of a category will be always used in the storefront. If the category contains featured products, this sort will be labelled "Featured", and this might prove confusing to some admins.  To resolve this, add a note just below the dropdown indicating what will occur when products are featured in the category.

### DIFF
--- a/admin/app/views/workarea/admin/catalog_categories/edit.html.haml
+++ b/admin/app/views/workarea/admin/catalog_categories/edit.html.haml
@@ -60,7 +60,7 @@
             .property
               = label_tag 'category_default_sort', t('workarea.admin.fields.default_sort'), class: 'property__name'
               = select_tag 'category[default_sort]', options_for_select(@category.sort_options, selected: @category.default_sort)
-              - if @category.product_ids.any?
+              - if @category.featured_products?
                 %span.property__note
                   = t('workarea.admin.catalog_categories.edit.default_sort_note')
 
--- a/admin/app/views/workarea/admin/catalog_categories/edit.html.haml
+++ b/admin/app/views/workarea/admin/catalog_categories/edit.html.haml
@@ -60,6 +60,9 @@
             .property
               = label_tag 'category_default_sort', t('workarea.admin.fields.default_sort'), class: 'property__name'
               = select_tag 'category[default_sort]', options_for_select(@category.sort_options, selected: @category.default_sort)
+              - if @category.product_ids.any?
+                %span.property__note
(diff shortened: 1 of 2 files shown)


admin/app/views/workarea/admin/catalog_categories/edit.html.haml
```diff
@@ -60,6 +60,8 @@
             .property
               = label_tag 'category_default_sort', t('workarea.admin.fields.default_sort'), class: 'property__name'
               = select_tag 'category[default_sort]', options_for_select(@category.sort_options, selected: @category.default_sort)
+              %span.property__note
```
bencrouse, line 63: You can check if there are any featured products here, so we only need to show them a note if that would actually affect this category.

admin/config/locales/en.yml
```diff
@@ -286,6 +286,7 @@ en:
             views: views
         edit:
           client_id_note: Identifies the category for product import
+          default_sort_note: 'When a category has featured products, this sort will appear as the "Featured" option on the storefront.'
```
bencrouse, line 289: If we only output this when there are featured products, this can be really specific and helpful like: "Since there are featured products, this will determine the sort of products showing after the featured products." Or something explicit like that.

bencrouse (changes requested):

admin/app/views/workarea/admin/catalog_categories/edit.html.haml
```diff
@@ -60,6 +60,9 @@
             .property
               = label_tag 'category_default_sort', t('workarea.admin.fields.default_sort'), class: 'property__name'
               = select_tag 'category[default_sort]', options_for_select(@category.sort_options, selected: @category.default_sort)
+              - if @category.product_ids.any?
```
bencrouse, line 63:
```suggestion
              - if @category.featured_products?
```

admin/config/locales/en.yml
```diff
@@ -286,6 +286,7 @@ en:
             views: views
         edit:
           client_id_note: Identifies the category for product import
+          default_sort_note: 'This category has featured products. When a category has featured products, this sort will appear as the "Featured" option on the storefront.'
```
bencrouse, line 289:
```suggestion
          default_sort_note: 'Since this category has featured products, this sort controls the products listed after the selected featured products.'
```

bencrouse (approved):

PR #545, opened 2020-10-19
Clean up search settings UI

bencrouse:
This UI really needed a little TLC.

### DIFF
--- a/admin/app/views/workarea/admin/search_settings/show.html.haml
+++ b/admin/app/views/workarea/admin/search_settings/show.html.haml
@@ -10,45 +10,62 @@
     = form_tag search_settings_path, method: :patch, data: { unsaved_changes: '' } do
       .tabs
         .tabs__panel
-          %h2.tabs__heading= t('workarea.admin.search_settings.show.synonyms.title')
-          %p= t('workarea.admin.search_settings.show.synonyms.description')
-          %p= t('workarea.admin.search_settings.show.synonyms.simple_example_description')
-          %pre
-            :preserve
-              #{t('workarea.admin.search_settings.show.synonyms.simple_example')}
-          %p= t('workarea.admin.search_settings.show.synonyms.phrase_example_description')
-          %pre
-            :preserve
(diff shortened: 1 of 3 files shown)


admin/app/views/workarea/admin/search_settings/show.html.haml
```diff
@@ -10,45 +10,62 @@
...
-                  = text_field_tag "boosts[#{field}]", value, class: 'text-box text-box--small'
+              .grid__cell.grid__cell--50.align-center
+                %h3= t('workarea.admin.search_settings.show.boosts.field_weight')
+                .grid
+                  .grid__cell.grid__cell--50
+                    %p= t('workarea.admin.search_settings.show.boosts.field_weight_description')
+                - @settings.boosts.each do |field, value|
+                  .property
+                    = label_tag "boosts[#{field}]", field.to_s.titleize, class: 'property__name'
+                    = range_field_tag "boosts[#{field}]", value, min: 0, max: 5, step: 0.1, class: 'text-box text-box--medium'
```
bencrouse, line 68: This all started with wanting to use these ranges to limit how far people adjust the boosts.

tubbo: Not sure if you meant to do this, but the `<label>` tags are a lot longer than the `<input>` tags they're for:

Screen Shot 2020-10-19 at 3 38 01 PM

I noticed this by accidentally clicking some whitespace to the right of the label tag, and saw one of the `<input>`s suddenly became selected. This could probably be avoided by ensuring the width of the label is the same as the width of the input (like by wrapping all these in a single `<div>` with a max-width or something). Not sure if it's a "dealbreaker" because I can't seem to accidentally change anything, so it might just look odd but not actually cause any problems...

bencrouse: I don't think that's a big deal, we can revisit later if it causes some grief.

mttdffy (approved):

tubbo (approved):

PR #546, opened 2020-10-21
Include Guest Orders In Admin For Users

tubbo:
When looking up recent orders for a user in the admin, also include any guest orders placed with the same email address. This fixes a discrepancy between the amount of orders seen in admin for a given user, and the amount of orders attributed to a user within insights, as insights treat all orders with the same email address as coming from the same user.

### DIFF
--- a/admin/app/view_models/workarea/admin/user_view_model.rb
+++ b/admin/app/view_models/workarea/admin/user_view_model.rb
@@ -57,7 +57,14 @@ module Workarea
 
       def orders
         return [] unless model.email.present?
-        @orders ||= OrderViewModel.wrap(Order.recent(model.id, 50))
+
+        @orders ||= OrderViewModel.wrap(
+          Order
+            .placed
+            .where(email: model.email)
+            .order_by([:placed_at, :desc])
+            .limit(50)
+        )
       end
 
       def insights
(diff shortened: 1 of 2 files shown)


admin/app/view_models/workarea/admin/user_view_model.rb
```diff
@@ -57,7 +57,14 @@ def payment_profile
 
       def orders
         return [] unless model.email.present?
-        @orders ||= OrderViewModel.wrap(Order.recent(model.id, 50))
+
+        @orders ||= OrderViewModel.wrap(
+          Order
+            .where(email: model.email, :placed_at.exists => true)
```
bencrouse, line 63: It would be better to reuse the `placed` scope, which will eliminate the need for this `exists` check and the `excludes` below.

bencrouse (changes requested):

admin/app/view_models/workarea/admin/user_view_model.rb
tubbo: Note that `Order.recent` does _not_ use the `:placed` scope, we might want to change that at some point in the future.

bencrouse (approved):

PR #547, opened 2020-10-28
Include User Orders When Email Address Changes

tubbo:
To protect against orders going missing when a user changes their email address, Workarea now queries for all orders with the email or ID of the user being displayed in `Admin::UserViewModel#orders`.


bencrouse:
Closing in favor of #548

PR #552, opened 2020-10-30
Fix Mongoid query cache data mutation

bencrouse:
Because the model shares the same underlying `BSON::Document` instance
as the query cache, when you change data on the model, the query cache
also changes. So if you do a find for the same model, you get a new
instance with mutated data which doesn't match the database.

This is particularly noticable as it relates to our releases
functionality, where documents loaded out of the database are
intentionally mutated, but not saved.


bencrouse:
One possible risk here, is the number of objects created in memory. Any better ideas?

tubbo (approved):

mttdffy (approved):

bencrouse:
closing because this has a pretty dramatic effect on memory:

Screen Shot 2020-11-04 at 3 28 00 PM

PR #555, opened 2020-11-03
Fix Missing Instance Variable In Cart Items View

tubbo:
The `@cart` instance variable was only being conditionally defined if `current_order.add_item` succeeded. This caused an error if `#add_item` happens to fail when calling `POST /cart/items` from the storefront, resulting in a 500 error. To prevent this error, the definition of this variable has been moved above the condition.

### DIFF
--- a/storefront/test/integration/workarea/storefront/cart_items_integration_test.rb
+++ b/storefront/test/integration/workarea/storefront/cart_items_integration_test.rb
@@ -47,6 +47,18 @@ module Workarea
         assert_equal(5.to_m, order.items.first.total_price)
       end
 
+      def test_fail_to_add_an_item
+        post storefront.cart_items_path,
+          params: {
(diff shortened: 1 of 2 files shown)


storefront/test/integration/workarea/storefront/cart_items_integration_test.rb
```diff
@@ -47,6 +47,30 @@ def test_can_add_an_item
         assert_equal(5.to_m, order.items.first.total_price)
       end
 
+      def test_fail_to_add_an_item
+        password = 'W3bl1nc!'
+        user = create_user(password: password)
+        order = create_order(user_id: user.id, email: user.email)
+
+        post storefront.login_path,
```
bencrouse, line 55: What does this logging in test?

bencrouse (approved):

PR #560, opened 2020-11-16
Add warning to inform developers why redirects aren't working locally

bencrouse:
This has confused developers a couple of times, so hopefully adding a
warning will help.

### DIFF
--- a/admin/config/locales/en.yml
+++ b/admin/config/locales/en.yml
@@ -2288,6 +2288,10 @@ en:
           destination_label: Redirect To
           destination_note: Can be any URL
           destination_placeholder: "/new/path.html"
+          disabled: >
+            This application is configured to consider requests to be local, so the Workarea exception handling isn't used.
+            This is most likely because it's running the development environment, where you want exceptions to be seen in responses.
+            These redirects won't work without this.
           from: From
           import: Import Redirects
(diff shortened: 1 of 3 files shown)


admin/config/locales/en.yml
```diff
@@ -2288,6 +2288,7 @@ en:
           destination_label: Redirect To
           destination_note: Can be any URL
           destination_placeholder: "/new/path.html"
+          disabled: This application is configured to consider requests to be local, so the Workarea exception handling isn't used. This is most likely because you are running the development environment. These redirects won't work without this.
```
mttdffy, line 2291: Putting myself in the shoe's a someone who doesn't know about this, I feel like this message comes across as if something is not configured that should be. While that would be the case if this was shown in a non-development environment, I feel like the message could use some clearer messaging that its not expected to work in development. Would it help if we threw a quick article on redirects into the docs and pointed them to it for more clarification?

mttdffy (changes requested):

admin/config/locales/en.yml
bencrouse: I updated, feel free to take a stab at it if you want.

mttdffy (approved):

PR #562, opened 2020-11-23
Update display of release changeset to handle large changesets

mttdffy:
WORKAREA-310

### DIFF
--- a/admin/app/helpers/workarea/admin/changesets_helper.rb
+++ b/admin/app/helpers/workarea/admin/changesets_helper.rb
@@ -2,13 +2,22 @@ module Workarea
   module Admin::ChangesetsHelper
     def changeset_icon(changeset, options = {})
       type = changeset.root.model_name.element
-      inline_svg(icon_path(type), options)
+      inline_svg(releasable_icon_path(type), options)
     end
 
-    private
+    def releaseable_icon(model, options = {})
+      type = model.model_name.element
+      inline_svg(releasable_icon_path(type), options)
+    end
 
-    def icon_path(type)
-      Workarea.config.releasable_icons[type.to_sym] || "workarea/admin/icons/#{type}.svg"
+    def releasable_icon_path(type)
+      default = 'workarea/admin/icons/release.svg'
+      return default unless type.present?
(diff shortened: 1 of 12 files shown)


admin/app/helpers/workarea/admin/changesets_helper.rb
```diff
@@ -2,12 +2,15 @@ module Workarea
...
 
-    private
+    def releaseable_icon(model, options = {})
+      type = model.model_name.element
+      inline_svg(releasable_icon_path(type), options)
+    end
 
-    def icon_path(type)
+    def releasable_icon_path(type)
       Workarea.config.releasable_icons[type.to_sym] || "workarea/admin/icons/#{type}.svg"
```
bencrouse, line 14: It'd be good to add a default here just in case.

core/app/models/workarea/release/changeset.rb
```diff
@@ -19,6 +19,29 @@ class Changeset
       index('changeset.product_ids' => 1)
       index('original.product_ids' => 1)
       index('releasable_type' => 1, 'releasable_id' => 1)
+      index(updated_at: 1)
+
+      def self.latest(limit = Workarea.config.per_page)
+        order(updated_at: :desc).limit(limit)
+      end
+
+      def self.summary(release_id)
```
bencrouse, line 28: So why this instead of `Search::AdminSearch.new(q: '*').facets.detect { |f| f.name == 'upcoming_changes' }`? It seems like the results page we'd land you on would more likely match if the summary was based on those results.

bencrouse (changes requested):

core/app/models/workarea/release/changeset.rb
mttdffy: what you suggested would give us a list of releases, not a list of releasable types. I guess I could do a search for the release id in upcoming_changes and look at the type facet. Hadn't really thought about it.

bencrouse (changes requested):
Would you add that default icon before I merge?

bencrouse (approved):

PR #563, opened 2020-11-24
Bugfix/workarea 309 localized active fields

mttdffy:
I ran a build off a branch of this branch with `localized_active_field = false` here: https://github.com/workarea-commerce/workarea/actions/runs/379858423

### DIFF
--- a/core/lib/workarea/configuration/localized_active_fields.rb
+++ b/core/lib/workarea/configuration/localized_active_fields.rb
@@ -14,7 +14,7 @@ module Workarea
         return if Workarea.config.localized_active_fields
 
         ::Mongoid.models.each do |klass|
-          if klass < Releasable
+          if klass < Releasable && klass.localized_fields['active'].present?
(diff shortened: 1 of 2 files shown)


bencrouse (commented):
Looks good, is this relevant for `v3.4-stable`?

mttdffy:
> Looks good, is this relevant for `v3.4-stable`?

I deemed it unnecessary given that this would come up from the beginning for anyone on v3.4-stable already and we haven't heard about until now. One could make the argument that upgrading 1 minor at a time you would hit this if you set it to false on v3.4, but that is something I thought of just now. Do you feel like its worth it for that reason?

mttdffy:
@bencrouse after looking at v3.4. The one test that fails doesn't exists, and the complication with segments isn't an issue since it doesn't have the core segmentation in this version. Think this is good to just merge into v3.5

bencrouse (approved):

PR #569, opened 2020-12-09
Add reCAPTHCHA support

bencrouse:
This change allows configuring reCAPTCHA v3 and/or v2 in the admin. It
is placed on critical site flows where we've noticed problems with bot
abuse.

While the primary intent is using reCAPTCHA v3, we allow v2 to be
configured to support challenging when a v3 verification fails.

### DIFF
--- a/storefront/app/controllers/workarea/storefront/recaptcha.rb
+++ b/storefront/app/controllers/workarea/storefront/recaptcha.rb
@@ -1,0 +1,63 @@
+module Workarea
+  module Storefront
+    module Recaptcha
+      extend ActiveSupport::Concern
+
+      included do
+        helper_method :recaptcha_type
+        helper_method :challenge_recaptcha?
+      end
+
+      def recaptcha_type
+        v3 = Workarea.config.recaptcha_v3_site_key.present? && Workarea.config.recaptcha_v3_secret_key.present?
+        v2 = Workarea.config.recaptcha_v2_site_key.present? && Workarea.config.recaptcha_v2_secret_key.present?
+
+        result = if v3 && v2
+          'v3_with_challenge'
+        elsif v3
+          'v3'
+        elsif v2
+          'v2'
+        else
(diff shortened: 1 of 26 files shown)


storefront/app/controllers/workarea/storefront/recaptcha.rb
```diff
@@ -0,0 +1,63 @@
...
+          'v3_with_challenge'
+        elsif v3
+          'v3'
+        elsif v2
+          'v2'
+        else
+          'none'
+        end
+
+        result.inquiry
```
mttdffy, line 25: fancy

storefront/test/integration/workarea/storefront/recaptcha_integration_test.rb
```diff
@@ -0,0 +1,133 @@
...
+        ::Recaptcha.configuration.skip_verify_env.delete('test')
+      end
+
+      def disable_recaptcha
+        ::Recaptcha.configuration.skip_verify_env << 'test'
+      end
+
+      def setup_v3
+        Workarea.config.recaptcha_v3_site_key = '6LdxWf0ZAAAAAPb40p0---l92ifZiWbNOPX6ujYv'
+        Workarea.config.recaptcha_v3_secret_key = '6LdxWf0ZAAAAAPsETRjMxjmv2NEKjvJQ6xrhaxLB'
```
mttdffy, line 19: are these keys that are ok to commit? like official test keys or something?

mttdffy (changes requested):
relevant build failure

storefront/test/integration/workarea/storefront/recaptcha_integration_test.rb
bencrouse: Just some test ones I generated for cassette recording, but may as well update to fake ones I guess.

mttdffy (approved):

PR #574, opened 2020-12-15
Release changeset updates

mttdffy:


admin/app/controllers/workarea/admin/create_release_undos_controller.rb
```diff
@@ -13,10 +13,15 @@ def create
...
         if @undo_release.save
-          @release.changesets.each do |changeset|
+          @release.changesets.limit(Workarea.config.per_page).each do |changeset|
             changeset.build_undo(release: @undo_release.model).save!
           end
 
+          BuildReleaseUndoChangesets.perform_async(
+            @undo_release.id,
+            @release.id
+          ) if @release.show_changeset_summary?
```
bencrouse, line 23: I see the logic reuse, but this method is tied to display so if it were to get decorated or something, changing this logic would definitely be a surprise.

```suggestion
          ) if @release.changeset_count > Workarea.config.per_page
```

core/app/workers/workarea/build_release_undo_changesets.rb
```diff
@@ -0,0 +1,23 @@
...
+
+      existing_changesets = undo_release.changesets.to_a
+      matching_changeset = ->(changeset, existing_changesets) do
+        existing_changesets.any? do |cs|
+          changeset.releasable_type == cs.releasable_type &&
+          changeset.releasable_id == cs.releasable_id
+        end
+      end
+
+      release.changesets.each do |changeset|
```
bencrouse, line 17:
```suggestion
      release.changesets.each_by(500) do |changeset|
```

bencrouse (changes requested):

bencrouse (approved):

PR #578, opened 2020-12-17
Add documentation for config.send_transactional_emails

mttdffy:
WORKAREA-317

### DIFF
--- a/docs/source/articles/system-emails.html.md
+++ b/docs/source/articles/system-emails.html.md
@@ -135,3 +135,34 @@ end
 ```
 
 [ActionMailer::TestHelper](https://api.rubyonrails.org/v5.2.3/classes/ActionMailer/TestHelper.html) has more information about the assertions you can use here.
+
+## Disable System Emails
+
+There are two configuration values that control the sending of email:
+
+- `Workarea.config.send_email`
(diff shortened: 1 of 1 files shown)


bencrouse (changes requested):
Would you add something about how the default value prevents non-admin mail from being sent in non-production environments? That confusion is the main driver for adding this documentation.

mttdffy:
> Would you add something about how the default value prevents non-admin mail from being sent in non-production environments? That confusion is the main driver for adding this documentation.

It doesn't do that? https://github.com/workarea-commerce/workarea/blob/master/core/lib/workarea/configuration.rb#L888

bencrouse:
Oh my bad, this is the wrong config value! This is the one causing the confusion: https://github.com/workarea-commerce/workarea/blob/master/core/lib/workarea/configuration.rb#L927

mttdffy:
> Oh my bad, this is the wrong config value! This is the one causing the confusion: https://github.com/workarea-commerce/workarea/blob/master/core/lib/workarea/configuration.rb#L927

Ah. I will add documentation for that.

bencrouse (changes requested):
Looks good, can you update for `v3.5-stable`?

bencrouse (approved):

PR #580, opened 2020-12-23
Improve image URL detection for dialog_buttons

jonmast:
This only worked if the URL had only one dot (before the file extension)
and no query parameters. Run it through the URL parser to provide more
robust handling.


bencrouse (approved):

bencrouse:
Thanks @jonmast !

PR #587, opened 2021-01-22
Upgrade gems

bencrouse:


core/lib/workarea/configuration/sidekiq.rb
```diff
@@ -40,10 +42,17 @@ def configure_workarea!
             size: pool_size,
             pool_timeout: pool_timeout
           }
+
+          # From the sidekiq-unique-jobs README
+          config.death_handlers << lambda do |job, ex|
```
mttdffy, line 47: *insert death eaters joke here*

core/app/models/workarea/pricing/discount/conditions/order_total.rb
```diff
@@ -25,7 +25,7 @@ module OrderTotal
           #
           # @return [Boolean]
           #
-          def order_total?
+          def use_order_total?
```
mttdffy, line 28: I assume this is to not step on the mongoid created method, yes?

mttdffy (approved):

core/app/models/workarea/pricing/discount/conditions/order_total.rb
bencrouse: Yeah, exactly

PR #597, opened 2021-07-01
docker-composer -v is no longer works for the recent version 2.0.0-be…

roseliux:
…ta.4

image


roseliux:
please review @bencrouse

bencrouse (commented):
Hi, this change looks good to me. Do you want to do this for `v3.5-stable` though, so it can be released as a patch?

PR #1093, opened 2026-03-17
WA-VERIFY-093: Harden find_or_create_by semantics for Mongoid 8

### LINKED ISSUE #1085: WA-VERIFY-093: Verify find_or_create_by/upsert semantics under Mongoid 8 (5 call sites)
## Summary

Mongoid 8 alters `find_or_create_by` / upsert semantics. The readiness assessment (PR #1075) identified **5 call sites** that must be verified to ensure idempotency and uniqueness expectations remain correct.

## Approach

```bash
# Find find_or_create_by usage
rg "find_or_create_by" core/app core/lib --glob='*.rb'
```

For each call site:
- Add tests for concurrency/idempotency where appropriate
- Ensure unique indexes/constraints align with behavior
- Adjust to safer patterns if required (document rationale)

## Acceptance Criteria
- [ ] All `find_or_create_by` call sites verified with tests (where practical)
- [ ] No duplicate-record scenarios introduced
- [ ] Tests pass under default appraisal
- [ ] Notes file documents any patterns changed

## Client Impact
None expected (prevents subtle data integrity regressions).

## Related
- Research: PR #1075 (WA-VERIFY-083 Mongoid 8 readiness assessment)
- Blocks: #690 (WA-RAILS7-004 Mongoid 8 upgrade)

kitcommerce:
Mongoid 8 can surface duplicate-key errors more readily when `find_or_create_by` races under concurrency. Harden Workarea core call sites to preserve idempotent behavior.

Changes:
- Add uniqueness-enforcing indexes for non-`_id` get-or-create patterns:
  - `Search::Settings`: unique index on `index`
  - `Content`: unique+sparse index on `name`; unique+sparse compound index on `contentable_type/contentable_id`
- Rescue duplicate key (`E11000`) races in `Search::Settings.current` and `Content.for` and fall back to `find_by`
- Add/extend unit tests for idempotency:
  - `Search::Settings.current`
  - `Content.for` (string + contentable)
  - `Pricing::Discount::FreeGift#apply` does not create duplicate `Pricing::Sku`
- Notes: `notes/WA-VERIFY-093-mongoid8-find-or-create-by.md`

Testing:
- `bundle exec ruby -Icore/test core/test/models/workarea/search/settings_test.rb`
- `bundle exec ruby -Icore/test core/test/models/workarea/content_test.rb`
- `bundle exec ruby -Icore/test core/test/models/workarea/pricing/discount/free_gift_test.rb`

Client impact:
- None expected

Closes #1085

### DIFF
--- a/core/app/models/workarea/content.rb
+++ b/core/app/models/workarea/content.rb
@@ -17,8 +17,9 @@ module Workarea
 
     belongs_to :contentable, polymorphic: true, optional: true, index: true
 
-    index({ name: 1 })
+    index({ name: 1 }, unique: true, sparse: true)
     index({ contentable_type: 1 })
+    index({ contentable_type: 1, contentable_id: 1 }, unique: true, sparse: true)
     index({ 'blocks._id' => 1 })
 
     embeds_many :blocks,
@@ -43,6 +44,21 @@ module Workarea
         find_or_create_by(name: object.titleize)
       else
         find_or_create_by(
+          contentable_id: object.try(:id),
+          contentable_type: object.try(:class)
+        )
+      end
+    rescue Mongo::Error::OperationFailure => e
+      # Under concurrency, unique indexes (name for system content, and
+      # contentable_type/contentable_id for contentable content) can raise a
+      # duplicate key error if another process inserts first. Fall back to
+      # reading.
+      raise unless e.message.include?('E11000')
+
+      if object.is_a?(String)
+        find_by(name: object.titleize)
+      else
+        find_by(
           contentable_id: object.try(:id),
           contentable_type: object.try(:class)
         )
--- a/core/app/models/workarea/search/settings.rb
+++ b/core/app/models/workarea/search/settings.rb
@@ -13,6 +13,8 @@ module Workarea
       field :range_facets, type: Hash, default: {}
       list_field :terms_facets
 
+      index({ index: 1 }, unique: true)
+
       around_update :update_indexes
 
       # The site-specific {Search::Settings} to use for the current request.
@@ -21,8 +23,14 @@ module Workarea
       # @return [Search::Settings]
       #
       def self.current
-        Thread.current[:current_search_settings] ||
+        Thread.current[:current_search_settings] ||= begin
           find_or_create_by(index: Elasticsearch::Document.current_index_prefix)
+        rescue Mongo::Error::OperationFailure => e
+          # Under concurrency, the unique index on :index can raise a duplicate
+          # key error if another process inserts first. Fall back to reading.
+          raise unless e.message.include?('E11000')
+          find_by(index: Elasticsearch::Document.current_index_prefix)
+        end
       end
 
       def self.current=(settings)
--- a/core/test/models/workarea/content_test.rb
+++ b/core/test/models/workarea/content_test.rb
@@ -24,5 +24,32 @@ module Workarea
       assert_equal(@content, Content.from_block(block.id))
       assert_equal(@content, Content.from_block(block.id.to_s))
     end
+
+    def test_for_string_is_idempotent
+      Content.create_indexes
+      Content.delete_all
+
+      first = Content.for('home_page')
+      second = Content.for('home_page')
+
+      assert_equal(first.id, second.id)
+      assert_equal(1, Content.where(name: 'Home Page').count)
+    end
+
+    def test_for_contentable_is_idempotent
+      Content.create_indexes
+      Content.delete_all
+
+      foo = Foo.create!(name: 'Foo content')
+
+      first = Content.for(foo)
+      second = Content.for(foo)
+
+      assert_equal(first.id, second.id)
+      assert_equal(
+        1,
+        Content.where(contentable_type: 'Workarea::ContentTest::Foo', contentable_id: foo.id).count
+      )
+    end
   end
 end
--- a/core/test/models/workarea/pricing/discount/free_gift_test.rb
+++ b/core/test/models/workarea/pricing/discount/free_gift_test.rb
@@ -59,6 +59,18 @@ module Workarea
           )
           assert(discount.catalog_qualifies?(order))
         end
+
+        def test_apply_does_not_create_duplicate_pricing_skus
+          Pricing::Sku.create_indexes
+          create_pricing_sku(id: 'GIFT', prices: [{ regular: 5.to_m }])
+
+          discount = FreeGift.new(sku: 'GIFT')
+          order = Pricing::Discount::Order.new(Workarea::Order.new)
+
+          2.times { discount.apply(order) }
+
+          assert_equal(1, Pricing::Sku.where(id: 'GIFT').count)
+        end
       end
     end
   end
--- a/core/test/models/workarea/search/settings_test.rb
+++ b/core/test/models/workarea/search/settings_test.rb
@@ -13,6 +13,19 @@ module Workarea
           settings.sanitized_synonyms
         )
       end
+
+      def test_current_is_idempotent
+        Settings.create_indexes
+        Thread.current[:current_search_settings] = nil
+
+        first = Settings.current
+        second = Settings.current
+
+        assert_equal(first.id, second.id)
+        assert_equal(1, Settings.where(index: first.index).count)
+      ensure
+        Thread.current[:current_search_settings] = nil
+      end
     end
   end
 end
--- a/notes/WA-VERIFY-093-mongoid8-find-or-create-by.md
+++ b/notes/WA-VERIFY-093-mongoid8-find-or-create-by.md
@@ -1,0 +1,34 @@
+# WA-VERIFY-093 — Mongoid 8: `find_or_create_by` / upsert semantics
+
+## Summary
+Mongoid 8 can surface duplicate-key errors more readily when `find_or_create_by` races under concurrency.
+
+Workarea core has 5 call sites relying on idempotent “get-or-create” behavior:
+
+- `Workarea::Metrics::User.save_order` (`_id` = email)
+- `Workarea::Search::Settings.current` (`index`)
+- `Workarea::Content.for(String)` (system content by `name`)
+- `Workarea::Content.for(contentable)` (by `contentable_type` + `contentable_id`)
+- `Workarea::Pricing::Discount::FreeGift#apply` (`Pricing::Sku` by `_id`)
+
+## Changes
+### Enforce uniqueness where required
+- `Workarea::Search::Settings`
+  - Added unique index on `index`
+
+- `Workarea::Content`
+  - Added unique + sparse index on `name` (system content)
+  - Added unique + sparse compound index on `contentable_type/contentable_id` (contentable content)
+
+> `sparse: true` avoids applying the unique constraint to documents that do not persist those fields (Mongoid omits nil fields).
+
+### Be resilient to duplicate-key races
+- `Search::Settings.current` and `Content.for` now rescue duplicate key errors (`E11000`) and fall back to `find_by`.
+
+This preserves public behavior while preventing intermittent errors during concurrent first-write scenarios.
+
+## Tests
+Added/extended tests asserting idempotency for:
+- `Search::Settings.current`
+- `Content.for` (string + contentable)
+- `Pricing::Discount::FreeGift#apply` (does not create extra `Pricing::Sku` records)


kitcommerce:
## Architecture Review

**Verdict:** PASS_WITH_NOTES (LOW)

Key notes:
- MEDIUM: Unique index creation may fail on existing duplicates; consider upgrade guidance / dedupe task.
- LOW: Rescue branching in Content.for is acceptable but coupled to method shape.
- LOW: Thread-local caching change uses ||=; considered safe.
- LOW: Consider adding idempotency test for Metrics::User.save_order in follow-up.

kitcommerce:
## Security Review

**Verdict:** PASS_WITH_NOTES (LOW)

Notes:
- Error code detection via `e.message.include?("E11000")` is somewhat fragile; prefer `e.code == 11000` when available.
- Otherwise, pattern fails safe (re-raises on non-match) and fallback is read-only.

kitcommerce:
## Rails Conventions Review

**Verdict:** CHANGES_REQUIRED (HIGH)

Blocking items:
- HIGH: `Search::Settings.current` caching should remain request-scoped (Thread.current / RequestStore), not process-global memoization.
- MEDIUM: Prefer checking duplicate-key by numeric error code vs message substring.
- MEDIUM: `find_by` fallback can return nil; consider `find_by!` or explicit guard for clearer failures.

(See reviewer JSON in dispatcher logs for full details.)

kitcommerce:
## Dispatcher Note (review payload correction)

The prior Rails-conventions review that flagged process-global memoization appears to have been based on an incomplete payload summary.

In the actual diff, `Search::Settings.current` uses `Thread.current[:current_search_settings] ||= ...`, which is request/thread-scoped (not class-level).

I'm re-running the Rails-conventions review with the full diff. Please disregard the earlier “process lifetime cache” blocking note.

kitcommerce:
## Review: WA-VERIFY-093 — Harden `find_or_create_by` for Mongoid 8

**Verdict: ✅ Approve** — solid, targeted fix with good test coverage. A few notes below, none blocking.

---

### What's good

**Index strategy is correct.** Promoting `{ name: 1 }` to `unique: true, sparse: true` is the right call for system content — `name` is nil on contentable records (Mongoid omits nil fields), so `sparse: true` correctly scopes the constraint to only the rows where it matters. Same reasoning for the compound `{ contentable_type, contentable_id }` index. The rationale is explained well in the notes file.

**`||=` fix in `Settings.current` is a meaningful correctness improvement, not just a style change.** The original `Thread.current[:foo] || find_or_create_by(...)` would re-execute `find_or_create_by` on every call as long as the cached value was falsy (e.g. `false`, `nil`, or an empty object). `||=` with a `begin/rescue` block caches only after a successful read-or-create, and the `ensure` in the test properly clears thread state — nice.

**E11000 rescue pattern is appropriate for this use case.** The check-then-act race in `find_or_create_by` is real, and optimistic write + rescue-and-read is the standard MongoDB idiom. Scoping the rescue to `E11000` specifically (rather than all `OperationFailure`) is exactly right.

**Test quality is good.** Idempotency is asserted on both the returned ID and the document count, which catches both the "returns same object" and "didn't insert a duplicate" cases. The `ensure` block on `test_current_is_idempotent` cleans up thread-local state correctly — many reviewers miss that.

---

### Observations / minor notes

**`FreeGift` call site (`Sku.find_or_create_by(id: sku)`) is `_id`-based — low risk but worth a comment in the notes.**  
`_id` in MongoDB has a built-in unique constraint, so a duplicate-key race here already resolves correctly at the driver level without any additional rescue. The new test is still a good regression guard, but the mechanism is different from the name/contentable cases. The notes file lists all 5 sites together; a one-liner clarifying "no rescue needed here because `_id` is inherently unique" would help the next reader.

**`Metrics::User.save_order` (`find_or_create_by(id: email)`) is also `_id`-based for the same reason** — no change needed there either, and it's correctly left untouched.

**Migration note for existing deployments:** Adding `unique: true` indexes to collections that may already have duplicates (pre-Mongoid 8 data, or test fixtures) will cause `create_indexes` / `ensure_indexes` to raise. Workarea's `rake workarea:db:mongoid:create_indexes` task is tolerant of this (it catches `OperationFailure`), but operators with data going back to Workarea 3.x may want a one-time dedup script before running index creation in production. Worth a note in the PR description or upgrade guide, even a brief one.

**`Content.for` rescue block repeats the lookup logic from the happy path** — both branches re-derive `name: object.titleize` or the `contentable_id/type` pair. That's fine at this scale; just something to keep in mind if `Content.for`'s logic gets more complex in future.

---

### Out of scope but visible

The `contentable_type: 1` single-field index (line 21) now coexists with the new compound `{ contentable_type: 1, contentable_id: 1 }` index. MongoDB can use the compound index for queries filtering on `contentable_type` alone (prefix rule), so the standalone `contentable_type: 1` index is now redundant. Not a correctness issue — just a minor space overhead. Could be cleaned up in a follow-on, or noted as a known artifact.

---

### Summary

The core changes are correct, minimal, and well-tested. The E11000 rescue is scoped tightly, the index choices are justified, and the test improvements are genuinely useful. The migration concern around pre-existing duplicates is the one thing I'd consider surfacing more explicitly before this lands in an upgrade guide.

kitcommerce:
## Simplicity Review ✅ PASS

```json
{
  "reviewer": "simplicity",
  "verdict": "PASS",
  "severity": null,
  "summary": "Changes are minimal and proportional — no over-engineering, unnecessary abstractions, or accidental complexity introduced.",
  "findings": []
}
```

---

The diff adds exactly what the problem requires and nothing more:

- **No new abstractions** — no protocols, generics, factories, or strategy objects
- **E11000 string match** is established MongoDB convention; the pattern  is idiomatic Ruby/Mongo and widely understood
- **Rescue fallback logic** in `Content.for` mirrors the existing lookup branches — no novel complexity introduced
- **`||=` assignment fix** in `Search::Settings.current` is a one-character correction that also tightens thread-local caching — clean and direct
- **Sparse unique indexes** are a direct, correct solution to the uniqueness enforcement requirement

Complexity added is proportional to the concurrency problem being solved. Nothing here is built for hypothetical future needs.

kitcommerce:
## ✅ rails-conventions — PASS

```json
{
  "reviewer": "rails-conventions",
  "verdict": "PASS",
  "severity": null,
  "summary": "Changes follow Rails/Mongoid conventions throughout: business logic lives in model class methods, fallbacks use idiomatic find_by, and the ||= memoization fix corrects a latent bug in Settings.current.",
  "findings": []
}
```

### Notes

- **Fat model placement** — both rescue/fallback blocks belong in model class methods, which is exactly where this logic should live. No controller contamination.
- **Finder idiom** — all fallbacks use `find_by` (not `where(...).first`). ✓
- **`||` → `||=` in `Settings.current`** — the original code was a memoization bug: the `||` expression returned the result but never stored it in `Thread.current[...]`, so the thread-local cache was never populated. The `||=` fix is correct and idiomatic Ruby. Consider this a bonus correctness fix.
- **`e.message.include?('E11000')`** — `Mongo::Error::OperationFailure` is a coarse exception class with no typed subclass for duplicate-key errors. String-matching on the error code is the accepted Mongoid ecosystem pattern here.
- No N+1s, no service-object overuse, no callback side-effects crossing model boundaries, no non-resourceful routes introduced.

---
*Automated review — wave 1 · rails-conventions · model: anthropic/claude-sonnet-4-6*

kitcommerce:
## Database Review

**Verdict: APPROVE** ✅

### Index Analysis

| Index | Correct | Notes |
|-------|---------|-------|
| `name: 1, unique+sparse` | ✅ | Sparse excludes null-name docs (contentable-based Content). Sound. |
| `contentable_type+id, unique+sparse` | ✅ | Compound sparse excludes docs where all fields null. Correct semantics. |
| `index: 1, unique` (Settings) | ✅ | No sparse needed — field always present. |

### Thread.current Cache Fix
The `||` → `||=` change is a **real bug fix**. The original code evaluated `find_or_create_by` on every call since `||` doesn't assign to the thread-local. Good catch.

### E11000 Rescue Pattern
Standard and correct for MongoDB upsert races under Mongoid 8 (which no longer uses atomic `$setOnInsert`). The fallback `find_by` after duplicate key error is the right approach.

### Recommendations (non-blocking)

1. **Migration safety (medium):** If production data has existing duplicates, `db:mongoid:create_indexes` will fail. Consider adding a rake task or migration note to deduplicate before index creation. Document in deployment notes.

2. **Error code vs string match (low):** `e.message.include?('E11000')` works but `e.code == 11000` would be more robust against message format changes. Non-blocking — the string check is a common pattern.

3. **Nil guard on fallback (low):** In the astronomically unlikely case where the winning insert is deleted between the E11000 and the retry `find_by`, the fallback returns `nil`. Consider whether callers handle this gracefully. Practically a non-issue.

### Summary
Solid hardening of concurrent `find_or_create_by` for Mongoid 8. Indexes are correct, sparse semantics are appropriate, the thread-local cache fix is a bonus. No N+1 concerns — fallback queries only fire on race conditions.

*Reviewed by: database reviewer (automated)*

kitcommerce:
## Rails-Security Review

**Reviewer:** rails-security | **Verdict:** ✅ APPROVE (minor advisory)

### Summary

This PR correctly hardens `find_or_create_by` race conditions by adding unique indexes and rescuing duplicate-key errors with a find fallback. The pattern is sound and widely used in MongoDB-backed Rails applications. No security vulnerabilities introduced.

### Findings

#### 1. E11000 Detection via String Matching — Low Risk / Advisory

```ruby
raise unless e.message.include?('E11000')
```

The PR checks for duplicate-key errors by matching the string `'E11000'` in the exception message. The Mongo Ruby driver's `OperationFailure` exposes an integer `code` attribute — checking `e.code == 11000` would be more robust against message format changes across MongoDB versions or driver localization.

**Risk:** Low. The `E11000` string has been stable across all MongoDB versions to date and is unlikely to change. However, `e.code == 11000` is the canonical approach and would be a one-line improvement.

**Suggested alternative:**
```ruby
raise unless e.code == 11000
```

#### 2. Race Condition Handling — Correct ✅

The rescue-and-fallback pattern is the standard solution for `find_or_create_by` races:
1. Attempt `find_or_create_by` (which does find then insert)
2. If another process inserts between find and insert → catch `E11000`
3. Fall back to `find_by` (the record now exists)

Both `Content.for` and `Settings.current` implement this correctly. The `find_by` fallback will find the record that the competing process inserted.

#### 3. Thread-Local Cache Fix — Correct ✅

`Settings.current` changed from `||` to `||=`, which fixes a subtle bug where the thread-local was never actually cached. Good catch.

#### 4. Sparse Index Usage — Correct ✅

`sparse: true` on `Content.name` and `Content.contentable_type/contentable_id` correctly avoids uniqueness violations for documents where those fields are nil (Mongoid omits nil fields by default). This is the right approach for optional-field uniqueness.

#### 5. No Auth/AuthZ Impact ✅

All changes are in internal model class methods (`Content.for`, `Settings.current`). No user-facing input flows into the rescue path. No authentication or authorization boundaries are affected.

#### 6. No Sensitive Data Exposure ✅

Non-E11000 errors are properly re-raised (`raise unless ...`), so unexpected database errors are not silently swallowed. No error details leak to end users.

#### 7. Error Swallowing Surface — Acceptable ✅

The rescue is scoped to `Mongo::Error::OperationFailure` (not a broad `rescue => e`), and only E11000 is caught. This is appropriately narrow.

### Verdict

```json
{
  "reviewer": "rails-security",
  "pr": 1093,
  "verdict": "APPROVE",
  "confidence": "high",
  "security_risk": "none",
  "findings": [
    {
      "severity": "advisory",
      "title": "E11000 string matching vs error code",
      "recommendation": "Consider e.code == 11000 instead of e.message.include?('E11000') for robustness"
    }
  ],
  "summary": "No security issues. Race condition handling is correct. Minor advisory on error detection method."
}
```

kitcommerce:
## Test-Quality Review

**PR:** #1093 — WA-VERIFY-093: Harden `find_or_create_by` semantics for Mongoid 8
**Reviewer role:** test-quality
**Verdict:** ⚠️ PASS_WITH_NOTES

---

### Summary

The PR adds valuable sequential idempotency tests that correctly verify the steady-state behavior of the hardened code paths. The implementation approach (unique indexes + E11000 rescue) is solid. However, the tests do not directly exercise the concurrent race condition that motivates the change—a structural gap worth documenting even if not blocking.

---

### Findings

#### ✅ Sequential idempotency tests are meaningful

The added tests (`test_for_string_is_idempotent`, `test_for_contentable_is_idempotent`, `test_current_is_idempotent`, `test_apply_does_not_create_duplicate_pricing_skus`) correctly:

- Call `create_indexes` before asserting to ensure the unique index is in place
- Invoke the target method twice sequentially and assert `same id, count == 1`
- Verify the non-concurrent happy path doesn't regress

Sequential idempotency tests are an appropriate baseline for a Ruby/Rails test suite where writing reproducible concurrent tests is genuinely difficult and often flaky in CI.

#### ⚠️ Race-condition path is not tested

The specific code added to rescue `Mongo::Error::OperationFailure` (E11000) is **never exercised** by the current tests. The rescue branch exists to handle a window where two processes simultaneously find no record, both attempt insert, and the second gets a duplicate-key error. The tests never:

1. Simulate an in-flight concurrent insert
2. Stub/inject a `Mongo::Error::OperationFailure` to force the rescue branch
3. Verify that the rescued result is the correct existing record

This means the rescue path — arguably the most critical new logic — has zero test coverage. If the `find_by` call in the rescue block has a bug (wrong field name, wrong value), it would go undetected.

**Recommendation:** Add at least one unit test that stubs the duplicate-key error on the `find_or_create_by` call and asserts the rescue path returns the correct document. Example sketch:

```ruby
def test_for_string_rescues_duplicate_key_error
  Content.create_indexes
  existing = Content.for('home_page')

  # Simulate another process winning the race
  dup_error = Mongo::Error::OperationFailure.new('E11000 duplicate key error ...')
  Content.stub(:find_or_create_by, ->(*) { raise dup_error }) do
    result = Content.for('home_page')
    assert_equal existing.id, result.id
  end
end
```

A similar stub test is warranted for `Search::Settings.current`.

#### ✅ Thread.current cleanup in `settings_test.rb`

`test_current_is_idempotent` correctly uses an `ensure` block to reset `Thread.current[:current_search_settings] = nil`. This prevents test pollution in the thread pool. Well done.

#### ⚠️ Missing `ensure` in `content_test.rb` variants

`test_for_string_is_idempotent` and `test_for_contentable_is_idempotent` call `Content.delete_all` at the top of the test (setup style) rather than in a teardown/ensure. If the test fails mid-run, leftover records from a failed `Content.for` call could persist and influence sibling tests. Prefer:

```ruby
def test_for_string_is_idempotent
  Content.create_indexes
  Content.delete_all

  begin
    first = Content.for('home_page')
    second = Content.for('home_page')
    assert_equal(first.id, second.id)
    assert_equal(1, Content.where(name: 'Home Page').count)
  ensure
    Content.delete_all
  end
end
```

or rely on the test framework's existing database-cleaning hook if one is present.

#### ℹ️ `FreeGift#apply` test: no index + limited scenario

`test_apply_does_not_create_duplicate_pricing_skus` calls `Pricing::Sku.create_indexes` and calls `discount.apply(order)` twice. This is fine for the sequential case. However:

- `FreeGift#apply` uses `find_or_create_by(id: sku_id)`, which relies on `_id` uniqueness (a MongoDB guarantee). No E11000 rescue was added for this site (per the notes file), and the test passes—but it doesn't add new safety net. The test is net-positive (regression guard) but low-signal for this PR's core concern.

#### ℹ️ `Metrics::User.save_order` call site — unaddressed

The notes file lists `Workarea::Metrics::User.save_order` as one of 5 call sites. It relies on `_id = email` uniqueness (also a MongoDB guarantee) and was intentionally left without changes. No test was added or modified for this call site. The rationale is reasonable (`_id` uniqueness is atomic), but a brief inline comment in the source noting the intentional skip would prevent future reviewers from wondering if it was overlooked.

---

### Verdict

| Dimension | Assessment |
|---|---|
| Sequential idempotency | ✅ Covered |
| Concurrent/E11000 rescue path | ❌ Not covered |
| Thread.current cleanup | ✅ Correct in settings_test |
| Test isolation (teardown) | ⚠️ Minor: content_test uses setup-style delete_all only |
| Edge cases | ℹ️ FreeGift low-signal; Metrics::User intentionally skipped |

**Overall: PASS_WITH_NOTES** — The tests are a solid baseline and the implementation logic is correct. The gap is that the rescue path (the novel hardening) is unexercised. This is a medium-severity gap: the code is correct, but confidence in the fallback path relies entirely on code review rather than tests. Recommend adding rescue-path stub tests before merge if feasible; otherwise document the gap.

---

*Reviewed by Kit (test-quality reviewer, wave 2)*

kitcommerce (commented):
Wave 2 review (7x)

```json
{"reviewer":"test-quality","verdict":"PASS_WITH_NOTES","severity":"LOW","summary":"Added idempotency tests are good, but they don't exercise true concurrency; still valuable regression coverage.","findings":[{"severity":"LOW","file":"core/test/models/workarea/content_test.rb","line":24,"issue":"Tests validate repeated calls, but not concurrent creation races (the production failure mode).","suggestion":"Optional: add a stress/concurrency test if the CI environment allows it; otherwise this is acceptable as a deterministic guard."}]}
```
```json
{"reviewer":"architecture","verdict":"PASS","severity":null,"summary":"Model-level idempotent-get-or-create behavior is appropriately encapsulated in the model APIs.","findings":[]}
```
```json
{"reviewer":"database","verdict":"PASS_WITH_NOTES","severity":"MEDIUM","summary":"Unique indexes + duplicate-key fallback is the right shape for Mongoid 8 races; error-class/message matching could be made more robust.","findings":[{"severity":"MEDIUM","file":"core/app/models/workarea/content.rb","line":48,"issue":"Rescuing Mongo::Error::OperationFailure and checking e.message.include?('E11000') is somewhat brittle across driver versions/locales.","suggestion":"Prefer checking an error code when available (e.g. e.respond_to?(:code) && e.code == 11000) or rescuing the more specific duplicate-key exception type if Mongoid/driver exposes it."},{"severity":"MEDIUM","file":"core/app/models/workarea/search/settings.rb","line":23,"issue":"Same brittle duplicate-key detection via message substring.","suggestion":"Use error code (11000) / label matching if supported to reduce false negatives/positives."},{"severity":"LOW","file":"core/app/models/workarea/content.rb","line":17,"issue":"Adding unique indexes will require index creation in existing deployments; duplicates will cause index build failures.","suggestion":"Consider noting in upgrade docs / release notes that indexes must be built and duplicates (if any) resolved before deploy."}]}
```
```json
{"reviewer":"security","verdict":"PASS","severity":null,"summary":"No new security exposure; uniqueness constraints can reduce unexpected data states.","findings":[]}
```
```json
{"reviewer":"rails-security","verdict":"PASS","severity":null,"summary":"No authn/authz or input handling changes; safe.","findings":[]}
```
```json
{"reviewer":"simplicity","verdict":"PASS","severity":null,"summary":"Straightforward: enforce uniqueness and fall back to read-on-duplicate.","findings":[]}
```
```json
{"reviewer":"rails-conventions","verdict":"PASS_WITH_NOTES","severity":"LOW","summary":"Pattern is consistent with Mongoid practice; minor style/readability tweaks possible.","findings":[{"severity":"LOW","file":"core/app/models/workarea/content.rb","line":57,"issue":"Inline `if object.is_a?(String) then ... else ... end` is valid but slightly harder to scan.","suggestion":"Consider a multi-line if/else without `then` for readability (non-blocking)."}]}
```

